PR #4, opened 2025-05-12
Instead of writing directly to CSV files, post to the Gradio server and have the UI write to a .sqlite db

abidlabs:


trackio/run.py
```diff
@@ -1,18 +1,25 @@
...
     ):
         self.project = project
+        self.client = client
         self.name = name or generate_readable_name()
         self.config = config or {}
-        self.storage = TrackioStorage(project, name, config)
 
     def log(self, metrics: dict):
-        self.storage.log(metrics)
+        self.client.predict(
```
znation, line 20: Is it called `predict` because this API was originally used to run model predictions? I guess it's more of a generic server-API-calling method now?

abidlabs: Exactly

PR #10, opened 2025-05-14
Move `trackio` dir to within `HF_HOME` and add some basic unit tests

abidlabs:


abidlabs:
Very small PR so I'll go ahead and merge this in, cc @znation for visibility.

Issue #7, opened 2025-05-13
Remove the flicker that happens when the lineplots are refreshed with more data


abidlabs:

abidlabs:
Yes, we need to avoid both @aliabid94

PR #21, opened 2025-05-19
Add range selection and keys

aliabid94:
Adds range selection (for zooming in and out) and keys (to prevent re-rendering flicker).

Keying is done naively - using the index of the plot. Can probably be done better to preserve matching of related plots

Closes: https://github.com/gradio-app/trackio/issues/7


abidlabs (approved):
works beautifully thanks @aliabid94!

PR #27, opened 2025-05-20
Auto-deploy to Spaces

znation:
* Add a space_id param to trackio.init, which will use an existing trackio space or deploy one on demand if it does not yet exist.
* Fix imports in ui.py and its relative dependencies to work both within and outside of a package/module. Gradio loads its app file outside of a module context.
* Add a README.md to trackio to default to the correct app file when uploading to a new Space.

TODO / not yet tested:

* Block metrics from non-authenticated clients (I think it is currently wide open and always deploys on a public space)
* Better UX for specifying HF_TOKEN when not specified

### DIFF
--- a/trackio/README.md
+++ b/trackio/README.md
@@ -1,0 +1,5 @@
+---
+sdk: gradio
+sdk_version: 5.30.0
+app_file: ui.py
+---
--- a/trackio/README.md
+++ b/trackio/README.md
@@ -1,5 +1,5 @@
 ---
 sdk: gradio
-sdk_version: GRADIO_VERSION
+sdk_version: {GRADIO_VERSION}
 app_file: ui.py
 ---
--- a/trackio/README.md
+++ b/trackio/README.md
@@ -1,5 +1,5 @@
 ---
 sdk: gradio
-sdk_version: 5.30.0
+sdk_version: GRADIO_VERSION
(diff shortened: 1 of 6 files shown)


znation:
@abidlabs if this looks good to you please approve and I'll merge. Thanks!

abidlabs:
Sorry @znation for the delay, will review this tom!

znation:
@abidlabs oh, no worries, I am not blocked so please take your time.

trackio/README.md
```diff
@@ -0,0 +1,5 @@
+---
+sdk: gradio
+sdk_version: 5.30.0
```
abidlabs, line 3: I think it'd be better to get the user's `gr.__version__` and use that here to ensure a consistent experience locally and on Spaces.

abidlabs:
Works great @znation! https://github.com/gradio-app/trackio/pull/27

image

Let me just add some validation logic & then we should be good to merge in

abidlabs (approved):
LGTM @znation! You may want to add a section in the trackio github repo README.md with this stuff!

trackio/README.md
abidlabs: updated

znation:
Thanks @abidlabs, I will add a README section and then merge.

PR #29, opened 2025-05-28
Optionally persist to a dataset

znation:
Uses `huggingface_hub.CommitScheduler` to automatically persist the
SQLite database to a dataset, if specified by the user. If dataset does
not exist it is automatically created.


trackio/__init__.py
```diff
@@ -34,6 +34,8 @@ def init(
     project: str,
     name: str | None = None,
     space_id: str | None = None,
+    persistent_dataset: str | None = None,
+    persistent_dataset_dir: str | None = None,
```
abidlabs, line 38: Do we need the `persistent_dataset_dir`? For simplicity for the user, can we exclude this and just choose a reasonable default path?

trackio/__init__.py
```diff
@@ -34,6 +34,8 @@ def init(
     project: str,
     name: str | None = None,
     space_id: str | None = None,
+    persistent_dataset: str | None = None,
```
abidlabs, line 37: Can we add docstrings for this param? And for consistency with `space_id`, perhaps we call this:
```suggestion
    dataset_id: str | None = None,
```
the persistence can be explained in the docstring

abidlabs (commented):
Looks very clean @znation! I left a few comments just to simplify the developer-facing experience. Let me know what you think.

znation:
Thanks for the improvements @abidlabs! Merged.

PR #30, opened 2025-06-04
Key event listeners and fix layout

aliabid94:
This adds keyed select and zoom event listeners in the render. Also fixed layout - instead of forcing two rows, just using a single row + min_width.


abidlabs (approved):
lgtm thanks @aliabid94! just updated the pinned version of `gradio` to 5.33

PR #31, opened 2025-06-04
Check HF_TOKEN when logging

znation:
To support the intended use case of a public Space that only the authorized users can write logs to (but anyone can view), pass in and check the ownership of the HF_TOKEN against the owner of the Space. Note: this only checks the ownership of the token against the ownership of the Space, and does not look at any specific permissions.

### DIFF
--- a/trackio/ui.py
+++ b/trackio/ui.py
@@ -2,7 +2,10 @@ import os
 from typing import Any
 
 import gradio as gr
+import huggingface_hub as hf
 import pandas as pd
+
+HfApi = hf.HfApi()
 
 try:
     from trackio.sqlite_storage import SQLiteStorage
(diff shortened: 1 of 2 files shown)


trackio/ui.py
```diff
@@ -142,9 +145,22 @@ def toggle_timer(cb_value):
...
+        if hf_token is None:
+            raise "Expected an hf_token to be provided when logging to a Space"
+        who = HfApi.whoami(hf_token)
+        owner_name = os.getenv("SPACE_AUTHOR_NAME")
+        # make sure the token user is either the author of the space,
+        # or is a member of an org that is the author.
+        # TODO: we should probably reject read-only or too-fine-grained tokens too...?
+        # but the logic to do that would be much more complex b/c of fine-grained tokens.
+        orgs = [o["name"] for o in who["orgs"]]
+        if owner_name != who["name"] and owner_name not in orgs:
```
abidlabs, line 162: I wonder if there's a built-in utility in `huggingface_hub` that handles this. cc @wauplin for your thoughts

abidlabs:
Nice @znation thanks for adding this check. Overall logic looks good, added a couple of minor suggestions

trackio/ui.py
hanouticelina: no there isn't a built-in utility in `huggingface_hub` that directly checks whether a token belongs to the author of a repo or a member of the owning org. comparing `who["name"]` against `SPACE_AUTHOR_NAME` is currently the right approach.
That said (and as mentioned by the TODO line 159), if you're using the token to write to the Space, it's not enough to just verify the username, you should also check the token  access level via `whoami(hf_token)["auth"]["accessToken"]["role"]`. It should be either `write` or `fineGrained`, and in the latter case, you would need to confirm it has write access to the specific repo or write permission to the owning org.

znation: Thank you @hanouticelina! I will leave it as a TODO in this PR but follow up with your advice in another PR.

znation: Actually I should just do it in this PR. Will add to this PR.

znation: It's too bad there isn't a write-access equivalent of (or parameter for) `auth_check`. Makes me wonder if we should add one but that goes deep (i.e. new moon-landing endpoint?).

Wauplin: > It's too bad there isn't a write-access equivalent of (or parameter for) auth_check. Makes me wonder if we should add one but that goes deep (i.e. new moon-landing endpoint?).

opened https://github.com/huggingface-internal/moon-landing/pull/14057 to tackle it

trackio/ui.py
```diff
@@ -139,9 +142,53 @@ def toggle_timer(cb_value):
         return gr.Timer(active=False)
 
 
-def log(project: str, run: str, metrics: dict[str, Any], dataset_id: str) -> None:
-    # Note: the type hint for dataset_id should be str | None but gr.api
+def log(
+    project: str, run: str, metrics: dict[str, Any], dataset_id: str, hf_token: str
+) -> None:
+    # Note: the type hint for dataset_id and hf_token should be str | None but gr.api
     # doesn't support that, see: https://github.com/gradio-app/gradio/issues/11175#issuecomment-2920203317
```
abidlabs, line 149: Unrelated, but in the latest version of `gradio`, we've fixed this, so now we can update the type hints to be `str | None`.

abidlabs (approved):
LGTM @znation nice PR!

abidlabs:
This looks great. We can bump the `__version__` so that when we merge in this PR, a new release of `trackio` will automatically happen, and I think that version should be ready for beta testing!

znation:
Bumped version to 0.0.12.

PR #35, opened 2025-06-06
Bug fix: key param not recognized

znation:
The `select` call doesn't appear to take a `key` param and gives an error. This commit removes the param which appears to fix the error.


abidlabs:
Hey @znation qq are you seeing this error with the latest version of Gradio?

znation:
Bug no longer repros w/ latest Gradio, closing.

PR #36, opened 2025-06-06
Queue up log items when Client is not ready

### LINKED ISSUE #28: Queue /logs until they can be called on the remote Space
Currently, when we connect to a remote Space, we wait until the Space is up, which is a blocking operation. As suggested by @pngwn, we should queue /log updates instead of blocking the script.

znation:
Instead of blocking while waiting for a Space to initialize, keep the client as None and attempt to initialize it lazily on each log call. When a client is successfully initialized, the log call will flush the queue.

Fixes #28

### DIFF
--- a/trackio/run.py
+++ b/trackio/run.py
@@ -1,3 +1,5 @@
+from collections import deque
+
 from gradio_client import Client
 
 from trackio.utils import generate_readable_name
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks @znation sorry I missed the thread earlier, but this approach you suggested sounds good to me

trackio/run.py
```diff
@@ -7,32 +12,78 @@
...
-        self.client = client
+        self._client_lock = threading.Lock()
+        self._client = client
         self.name = name or generate_readable_name()
         self.config = config or {}
         self.dataset_id = dataset_id
+        self._queued_logs = deque()
+
+        if client is None:
+            # kick off a thread to initialize the client
```
abidlabs, line 32:
```suggestion
```

abidlabs:
Hmm I'm seeing a 

```
[trackio] Failed to initialize client in background: can't create new thread at interpreter shutdown
```

error when running e.g. `python examples/deploy-on-spaces.py`

znation:
Thanks @abidlabs, I think with [this commit](https://github.com/gradio-app/trackio/pull/36/commits/3ad7dfeee2ca9091528e2ebff226d8598cc7f72c) it should work from both a script and Jupyter notebook. In order to avoid errors and data loss we have to block at some point and make sure the client is instantiated, so I have done this on the `finish` call to make sure the queue flushes by the end of `finish`.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "pandas",
     "huggingface-hub",
     "gradio>=5.34.2",
+    "backoff>=2.2.1",
```
abidlabs, line 17: Instead of adding a new dependency, could we just vendor in the code? my understanding is that we are just using this for this generator:

```py
def fibo():
    """Generator for Fibonacci backoff: 1, 1, 2, 3, 5, 8, ...
    """
    a, b = 1, 1
    while True:
        yield a
        a, b = b, a + b
```

abidlabs:
Nice @znation! Code looks good I tested with , and it didn't block the code execution while waiting for the Space to start.

The only thing I noticed was that it kept printing `[trackio] Failed to initialize client in background: The read operation timed out` every couple of seconds, while the Space was starting.

In the end, my console looked like this:

```
[trackio] Failed to initialize client in background: The read operation timed out
[trackio] Failed to initialize client in background: The read operation timed out
[trackio] Failed to initialize client in background: The read operation timed out
[trackio] Failed to initialize client in background: The read operation timed out
[trackio] Successfully initialized client in background.
```

imo it'd be better to hide these. otherwise, lgtm

abidlabs:
Just a heads up that `trackio.import_csv` now includes these two calls as well:

```
        deploy.create_space_if_not_exists(space_id, dataset_id)
        deploy.upload_db_to_space(project, space_id)
```

which we should queue

PR #40, opened 2025-06-12
Misc fixes

abidlabs:
Several fixes to the Gradio app:

* All runs should be loaded initially when a project is selected instead of just the first one
* Use a rolling window smoothing function instead of exponential averaging to avoid skewing to initial values
* Actually refresh the plot in realtime for multi-run projects (previously, it would only work for the first run)

Documentation fixes
* Better initial example in `fake-training`
* Link examples from README

Side note: if `trackio` picks up steam, we should add some playwright-based tests to avoid regressions


abidlabs:
@znation I'll merge this in to get it out to our beta testers because the refresh bug is quite frustrating actually

znation (commented):
Looks good to me!

PR #48, opened 2025-06-17
Track system metrics

abidlabs:


abidlabs:
Closing as I agree with @qgallouedec that we should let the user track their own metrics, including system metrics, and keep trackio focused and minimal

Issue #55, opened 2025-06-19
Allow importing logs from csv


abidlabs:
As mentioned by @NouamaneTazi:

* Would be nice to be able to import logs to trackio

Issue #58, opened 2025-06-19
Import static TensorBoard event traces (tfevent)


znation:
From the tfevent files here (for instance): https://huggingface.co/aubmindlab/bert-base-arabertv02/tree/main/runs

We should be able to import them and create an equivalent dashboard to https://huggingface.co/aubmindlab/bert-base-arabertv02/tensorboard

PR #62, opened 2025-06-20
Allow importing Trackio dashboard from a CSV

abidlabs:
Closes: https://github.com/gradio-app/trackio/issues/55

e.g. run `python examples/import-csv.py` and `python examples/import-csv-to-spaces.py`

Also, does some cleanup. Previously, we had a bug, which is that a new instance of `CommitScheduler` would be created every time a metric was logged. I've refactored by keeping a single instance of `CommitScheduler` as a contextvar and refactored the `SqliteStorage` class to be just a set of static methods to avoid similar bugs in the future. I also removed `dataset_id` from the `log()` call as it does not need to be passed around since it can be read directly from the environment variable.

### DIFF
--- a/trackio/deploy.py
+++ b/trackio/deploy.py
@@ -133,6 +133,6 @@ def upload_db_to_space(project: str, space_id: str) -> None:
     client.predict(
         api_name="/upload_db_to_space",
         project=project,
-        db=handle_file(db_path),
+        uploaded_db=handle_file(db_path),
         hf_token=huggingface_hub.utils.get_token(),
     )
--- a/trackio/ui.py
+++ b/trackio/ui.py
@@ -247,15 +247,17 @@ def check_auth(hf_token: str | None) -> None:
             )
 
 
-def upload_db_to_space(project: str, db: gr.FileData, hf_token: str | None) -> None:
+def upload_db_to_space(
+    project: str, uploaded_db: gr.FileData, hf_token: str | None
+) -> None:
     check_auth(hf_token)
-    if os.path.exists(db["path"]):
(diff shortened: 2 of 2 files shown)


trackio/__init__.py
```diff
@@ -188,4 +131,104 @@ def show(project: str | None = None):
     dashboard_url = base_url + f"?project={project}" if project else base_url
     print(f"* Trackio UI launched at: {dashboard_url}")
     webbrowser.open(dashboard_url)
-    block_except_in_notebook()
+    utils.block_except_in_notebook()
+
+
+def import_csv(
```
znation, line 137: I wonder if we should put this in its own file `import.py` -- especially if we eventually do #58 as well.

znation (approved):
Looks great! I am impressed how little code is needed to ingest the CSV format.

abidlabs:
Thanks @znation for reviewing this PR! I fixed up the tests so it should be ready to go. I'll try to test some more tomorrow & see if can add an end-to-end test that tests the flow from importing a csv all the way to deployment to Spaces*

@NouamaneTazi let me know if this PR works for you! One known issue is that if there are too many points in a plot (>1,000), the performance of the trackio dashboard suffers, especially as you drag your mouse over the plots. We can fix that by adding the ability to subsample in Gradio's plots, but that will take some time to implement

abidlabs:
*For the end-to-end tests to run in CI, we will probably need a fine-grained HF token to work with trackio @znation

abidlabs:
Actually sorry hold off on testing, couple of todos left

PR #73, opened 2025-07-01
Fix Spaces issue

abidlabs:
Deploying to new Spaces didn't work if just the `space_id` without the username was specified because the wrong `space_id` was being set as the `url` which meant the Client was never able to connect. 

Test by running: `examples/deploy-on-spaces.py`

Now that we support fine-grained tokens, I'll try to add an e2e test that actually deploys Trackio to Spaces and confirms that the data has been successfully logged.

### DIFF
--- a/examples/persist-dataset.py
+++ b/examples/persist-dataset.py
@@ -1,57 +1,87 @@
+import math
 import random
 import time
 
-from tqdm import tqdm
-
 import trackio as wandb
 
-wandb.init(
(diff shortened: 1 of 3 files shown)


znation (approved):
Thanks for fixing that! LGTM

abidlabs:
Thanks @znation! I'll actually merge this in and add the e2e test on spaces later

PR #75, opened 2025-07-02
Include the Manifest.ini file within pyproject.toml

ParagEkbote:
We can include the manifest.ini file into pyproject.toml without any breaking changes, leading to one less config file. I've tested the command with `hatch build` and confirmed its working.

Could you please review?

cc: @abidlabs


abidlabs:
Thanks @ParagEkbote, I'm OK with this change.

Issue #76, opened 2025-07-02
update y-lim when we filter on a plot


NouamaneTazi:
In the image below, after filtering grad_norm the y scale still goes until 8
Would be nice to adjust it automatically to the values shown
<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/3d9368ee-eebe-48f9-b5bb-38a0cc74e5e2" />

abidlabs:
Should be closed via https://github.com/gradio-app/trackio/pull/77

PR #77, opened 2025-07-02
Downsample in the backend for better performance

aliabid94:
Right now, trackio gets quite glitchy at 100k+ steps because of all the data being sent back and forth. This should improve performance by downsampling on the backend.

### DIFF
--- a/trackio/ui.py
+++ b/trackio/ui.py
@@ -10,10 +10,10 @@ HfApi = hf.HfApi()
 
 try:
     from trackio.sqlite_storage import SQLiteStorage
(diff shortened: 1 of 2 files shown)


abidlabs:
Does this fix https://github.com/gradio-app/trackio/issues/76 btw?

abidlabs (approved):
discussed with @aliabid94 and this seems like a good perf improvement that will improve intial load time as well as time for zoom-ins etc.

@NouamaneTazi could you test to see if it fixes the y lim issue you saw

PR #81, opened 2025-07-11
Fixing some nits

abidlabs:


abidlabs:
Going to merge these in as they are very minor

PR #82, opened 2025-07-16
Fix multiple CommitSchedulers

znation:
The ContextVar is thread-local (one per thread), so using that as the source of what should be a single global CommitScheduler was the source of this bug. This commit moves the CommitScheduler to a class variable protected by a thread lock, and there is now a single global CommitScheduler.

This should fix any 429/500 errors from commits.

### DIFF
--- a/trackio/context_vars.py
+++ b/trackio/context_vars.py
@@ -3,7 +3,6 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from trackio.run import Run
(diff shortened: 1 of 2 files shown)


abidlabs (approved):
LGTM nice fix @znation!

I did see this error after I terminated the script:

>     raise RuntimeError('cannot schedule new futures after '
RuntimeError: cannot schedule new futures after interpreter shutdown

but I think that's expected and won't happen once we restrict syncing to a dataset from Spaces

PR #84, opened 2025-07-16
Import from dataset on launch

### LINKED ISSUE #79: Pull from HF dataset before pushing to it
Currently, we upload the trackio database for a project to a dataset but I think we don't ever read it back. So in the case your data is lost locally or Space restarts, the user wouldn't see their data recovered. (Unless CommitScheduler handles this internally?)

znation:
If there is an existing database in the dataset referenced by the TRACKIO_DATASET_ID env var, then copy that to the local DB path on first initialization of SQLiteStorage.

Fixes #79

### DIFF
--- a/trackio/sqlite_storage.py
+++ b/trackio/sqlite_storage.py
@@ -1,10 +1,12 @@
 import json
 import os
+import shutil
 import sqlite3
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


abidlabs (approved):
Makes sense, awesome!

Issue #83, opened 2025-07-16
Write to dataset as parquet


znation:
For efficiency reasons (it dedupes much better with [parquet](https://huggingface.co/blog/improve_parquet_dedupe) on HF), we should write to the dataset as parquet instead of sqlite.

PR #85, opened 2025-07-17
Take a drop of CommitScheduler from huggingface_hub

znation:
Takes a copy of commit_scheduler.py from commit
d0a948fc2a32ed6e557042a95ef3e4af97ec4a7c of huggingface/huggingface_hub.

This has two purposes:
1. Easier debugging/changing of the internals
2. Need to add a hook for on_before_push for the implementation of #83


abidlabs (approved):
Nice lgtm

PR #88, opened 2025-07-17
Persist to Dataset as parquet instead of sqlite

### LINKED ISSUE #83: Write to dataset as parquet

znation:
* Add on_before_commit to CommitScheduler
* Implement SQLite -> Parquet using Pandas as the go-between

Fixes #83

### DIFF
--- a/trackio/sqlite_storage.py
+++ b/trackio/sqlite_storage.py
@@ -81,9 +81,7 @@ class SQLiteStorage:
         Exports all projects' DB files as Parquet under the same path but with extension ".parquet".
         """
         # don't attempt to export (potentially wrong/blank) data before importing for the first time
-        print("Zach was here, in export")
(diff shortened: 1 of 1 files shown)


abidlabs (approved):
LGTM, code changes looks great @znation! 

To test this PR, I ran the `examples/persist-dataset.py` script, which created this Space: https://huggingface.co/spaces/abidlabs/trackio-test-space-819590, and this Dataset: 

which now includes a nice Dataset preview too!

image

My only suggestion would be to perhaps add unit tests for `export_to_parquet()` and `import_from_parquet()`

znation:
Thanks for the review @abidlabs! I'll add unit tests in another PR so as to expedite this one.

PR #89, opened 2025-07-18
Fix dataset loading properly

### LINKED ISSUE #79: Pull from HF dataset before pushing to it

znation:
The previous fix for #79 did not populate the projects list, and potentially had a race between whether the import would attempt first (w/o a projects list, meaning import would not succeed) and project list creation (from client init).

This change reads the project list from the dataset after importing, so the order of operations is always:

1. dataset import
2. populate projects list
3. first init & log call from client

This seems to fix #79 in a more robust way.

### DIFF
--- a/trackio/sqlite_storage.py
+++ b/trackio/sqlite_storage.py
@@ -1,6 +1,5 @@
 import json
 import os
-import shutil
 import sqlite3
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


trackio/sqlite_storage.py
```diff
@@ -233,11 +220,38 @@ def get_metrics(project: str, run: str) -> list[dict]:
...
+            with SQLiteStorage.get_scheduler().lock:
+                try:
+                    files = hfapi.list_repo_files(dataset_id, repo_type="dataset")
+                    for file in files:
+                        if not file.endswith(".db"):
+                            continue
+                        downloaded_path = hf.hf_hub_download(
+                            dataset_id, file, repo_type="dataset"
+                        )
+                        shutil.copy(downloaded_path, TRACKIO_DIR)
```
abidlabs, line 240: Instead of downloading and then copying, could we use the `local_dir` argument of `hf_hub_download`?

znation: Good call.

PR #95, opened 2025-07-26
Refactor `generate_readable_name` to ensure unique name generation

qgallouedec:
Typically, training scripts explicitly set a random seed, for example:

```python
np.random.seed(42)
```

This means that even with `resume="never"`, the run will restart with the exact same conditions — effectively repeating the same run over and over.

The updated name-generation method solves this by checking whether a name is already in use before assigning it.

Another modification is that the number are sequential (instead of random), which I think helps to remember which run is the newest.


abidlabs (approved):
Nice makes sense! Code changes lgtm, would be great if we can just unit test this

PR #106, opened 2025-07-30
Flatten metrics column in parquet representation

### LINKED ISSUE #105: Break out metrics JSON into multiple parquet columns
Instead of a single parquet column of type JSON, it would be nice if the parquet files had columns for each metric.

This can potentially be done at parquet conversion (import/export) time since it's more convenient to keep the JSON `metrics` column in SQLite.

znation:
* When exporting to parquet, flatten the `metrics` column into multiple columns of Parquet.
* When importing, do the reverse (if the parquet file doesn't already have a `metrics` column).
* If the parquet file has a metrics column (legacy format), it will skip the unflattening on import, so it should continue to work.
* Adds "metrics" to the reserved metric names so that it's not possible to log a metric name that will get this logic confused.

Fixes #105

### DIFF
--- a/trackio/sqlite_storage.py
+++ b/trackio/sqlite_storage.py
@@ -93,6 +93,14 @@ class SQLiteStorage:
             ):
                 with sqlite3.connect(db_path) as conn:
                     df = pd.read_sql("SELECT * from metrics", conn)
+                # break out the single JSON metrics column into individual columns
+                metrics = df["metrics"].copy()
+                metrics = pd.DataFrame(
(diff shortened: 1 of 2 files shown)


abidlabs (approved):
I think we should add python unit tests for the import and export methods now, but otherwise looks good to me!

Tested with this Trackio dashboard: https://huggingface.co/spaces/abidlabs/trackio-99463
This is the dataset: https://huggingface.co/datasets/abidlabs/trackio-99463_dataset

image

PR #108, opened 2025-07-30
Update `trackio` project description

abidlabs:


znation (approved):
LGTM! Thanks for fixing. I didn't realize this was in the project itself, I thought it would be in the PyPI web UI for some reason.

PR #110, opened 2025-07-31
Explicitly mention python version in readme

abidlabs:


znation (changes requested):
Should we also put this in pyproject.toml to prevent installation on older versions? I think something like [this](https://stackoverflow.com/a/72462723) should work.

PR #116, opened 2025-08-06
Remove `tbparse` as a required dependency

abidlabs:
Sorry should have caught this earlier, but we shouldn't include `tbparse` as a required dependency as it is quite heavy (e.g. it has `tensorflow` as a dependency). Instead, let's make this an extras in the package. Also cleans some of the testing code.


pyproject.toml
```diff
@@ -33,6 +32,9 @@ dev = [
     "pytest",
     "ruff==0.9.3"
 ]
+tensorboard = [
+    "tbparse==0.0.9"
```
abidlabs, line 36: pin because this package is not super heavily maintained and we don't want want a downstream breakage to break `trackio`

znation (approved):

Issue #125, opened 2025-08-07
Bug when resuming a run from previous step


universuen:
Hi! Trackio is a wonderful work for the community. Besides its nice features, I found a small bug when resume a run from previous step.

Assuming the program crashed at step T+X and we resume the run from checkpoint step T, trackio will continue to log with step T+X even if `step` is set explicitly in the `trackio.log` method.

E.g.,
```python
import trackio


if __name__ == "__main__":
    trackio.init(
        project="playground", 
        resume="allow",
        name="demo"
    )
    for step in range(100):
        trackio.log(
            {"data": step},
            step=step,
        )
        if step == 15:  # The run crashed at step 15
            print("Crash!")
            break

    # Now we resume the run from checkpoint at step 10
    trackio.init(
        project="playground", 
        resume="allow",
        name="demo",
    )
    for step in range(10, 100):
        trackio.log(
            {"data": step},
            step=step,
        )
    
    trackio.finish()
    trackio.show("playground")

```

The expected plot is a straight line from (0, 0) to (100, 100), but the final result is a poly line ending with (105, 100).

<img width="2498" height="466" alt="Image" src="https://github.com/user-attachments/assets/de91a6d7-5b57-4ea4-b461-acd29e9bb2a0" />

lewtun:
More generally, it would be great is repeated runs where a run name is not explicitly set can produce independent curves. Currently if one runs the code below twice in a row, the resulting runs are continued from each other which is confusing:

<img width="1190" height="598" alt="Image" src="https://github.com/user-attachments/assets/be911242-87f1-4781-8c6b-68ef0c77d5a7" />

```python
import os
from datasets import load_dataset

from trl import SFTTrainer, SFTConfig
from datasets import load_dataset

MODEL_ID = "HuggingFaceTB/SmolLM2-135M-Instruct"
MODEL_NAME = MODEL_ID.split("/")[-1]

os.environ["TRACKIO_PROJECT"] = "trackio-testing-v1"
os.environ["TRACKIO_SPACE_ID"] = "trl-sft-trackio"



dataset = load_dataset("trl-lib/Capybara", split="train[:1000]")
dataset = dataset.train_test_split(test_size=0.1)

args = SFTConfig(
    learning_rate=2.0e-5,
    max_steps=10,
    packing=True,
    per_device_train_batch_size=2,
    gradient_accumulation_steps=8,
    gradient_checkpointing=True,
    max_length=32,
    logging_steps=1,
[...]

PR #126, opened 2025-08-07
bugfix for issue #125

universuen:
bugfix for issue https://github.com/gradio-app/trackio/issues/125


znation (approved):

abidlabs (approved):
Thanks @universuen for catching this and the contribution! I added a test to prevent regressions & we'll do a patch release with this fix.

PR #131, opened 2025-08-09
Implement trackio.Table

### LINKED ISSUE #114: Support for logging tables
Similar to #100, but for table data (data arranged in rows/columns like a DataFrame).

znation:
Adds support for logging tables and showing the latest version of the table in the dashboard. Mimics the wandb.Table API but doesn't implement most parameters yet.

Fixes #114


znation:
Actually as I am thinking about it, storing tables more efficiently maybe should block this PR, since otherwise we'll have to think about the back-compat implications.

trackio/__init__.py
```diff
@@ -7,12 +7,15 @@
...
 from trackio.run import Run
 from trackio.sqlite_storage import SQLiteStorage
 from trackio.ui import demo
 from trackio.utils import TRACKIO_DIR, TRACKIO_LOGO_DIR
 
+Table = table.Table
+
 __version__ = Path(__file__).parent.joinpath("version.txt").read_text().strip()
 
 __all__ = ["init", "log", "finish", "show", "import_csv", "import_tf_events"]
```
abidlabs, line 21:
```suggestion
from trackio.data_types.table import Table
from trackio.imports import import_csv, import_tf_events
from trackio.run import Run
from trackio.sqlite_storage import SQLiteStorage
from trackio.ui import demo
from trackio.utils import TRACKIO_DIR, TRACKIO_LOGO_DIR

Table = table.Table

__version__ = Path(__file__).parent.joinpath("version.txt").read_text().strip()

__all__ = ["init", "log", "finish", "show", "import_csv", "import_tf_events", "Table"]
```

trackio/utils.py
```diff
@@ -8,7 +8,16 @@
...
-RESERVED_KEYS = ["project", "run", "timestamp", "step", "time", "metrics"]
+RESERVED_KEYS = [
+    "project",
+    "run",
+    "timestamp",
+    "step",
+    "time",
+    "metrics",
+    "data_type",
+    "x_axis",
```
abidlabs, line 19: why do we need to add these two?

examples/log-table.py
abidlabs, line 1: Might be good to put `log-table` and `mnist` inside a folder, also with a `requirements.file` just for better organization

abidlabs:
Nice script @znation! I ran it and the training finished successfully: 

image

but I don't see any tables / dataframes. Am I missing something?

Saba9:
> @abidlabs I think this is now ready for review again, with a better-motivating example this time (MNIST classifier).
> 
> @Saba9 I would like to use `trackio.Image` support for this example eventually but we don't need to coordinate on that yet. I would appreciate if you would take a quick look at this PR just to see if there are any glaring inconsistencies with what you had in mind, though.

@znation Looks good! I don't see any obvious inconsistencies. One thing we'd need to do to add image support is expand the table API to support adding images to the table. For example with wandb you'd do something like:

```
img = wandb.Image("image.png")
table = wandb.Table(columns=["loss", "output_img"])
table.add_data(0.20, img)
```

lewtun:
Thanks for this feature @znation ! It would be good to test how it looks with text content as this is the most common use in reinforcement learning, where one wants to track e.g. `(prompt, completion)` from the model over training

trackio/utils.py
znation: They are used as columns in the DataFrame, and get picked up as `object` type in Pandas. This is curious, and I will investigate a bit more, because now that you mention it I think they should be string type.

lewtun:
Can confirm this works for text as well! Perhaps it's more of a Gradio question, but is there some way to wrap the text in the table columns? Right now, the text is very hard to read because it doesn't wrap (see video)

Screenshot 2025-08-28 at 15 15 54

https://github.com/user-attachments/assets/01a79795-950d-462a-b3e0-a8f9aa4eb94e

abidlabs:
Yes let's try `wrap=True` @znation

lewtun:
I re-tested the `Table` feature with wrapping and it works well (see video)! 

https://github.com/user-attachments/assets/56ee517e-8e0e-4b9a-9939-b3bd8ce3b881

One thing I noticed with long texts (typical with reasoning models) is that we get:

* Very long cells due to wrapping that are hard to read because new lines don't seem to be formatted (see screenshot comparison with WandB below)
* The table seems to hang a bit on the rendering (see above video)

Screenshot 2025-09-02 at 11 25 44
Screenshot 2025-09-02 at 11 25 05

Although out of scope for this PR, I quite like the WandB interface for tables, which truncates the rows by default for fast navigation and then allows one to expand a row with either a preview or full pop-up (see video below). Adding such a feature would presumably require changes on the Gradio side, so I can open an issue there if that's preferred.

https://github.com/user-attachments/assets/2e456a79-2bdd-47ef-ad9c-db78a18ddd84

In any case, shall we merge this PR and iterate in follow-ups?

znation:
@lewtun sounds good to me!

@abidlabs if the code change looks good to you please approve (or feel free to merge), thanks!

qgallouedec:
@znation I added some documentation

abidlabs (approved):
Aweome PR @znation @qgallouedec @lewtun! LGTM. I'll go ahead and merge this and then do a release to get this out.

Issue #143, opened 2025-08-18
Deploy to space feature is broken


Saba9:
Noticed that I wasn't able to successfully deploy trackio to an HF spaces after pulling latest main. I tried narrowing it down and the issue appears to have started with #135. I tested by running `examples/deploy-to-space.py`. A space and dataset are created, however, neither ever displays nor receives data.

PR #146, opened 2025-08-19
Fix deployment to Spaces

abidlabs:
Fixes the issue from https://github.com/gradio-app/trackio/pull/135 which broke deployment to Spaces.

Closes: https://github.com/gradio-app/trackio/issues/143


znation (approved):
LGTM, thanks for fixing this!

PR #149, opened 2025-08-19
Group metrics by prefix (and even subprefix) in UI

abidlabs:
Currently, metrics with different prefixes such as "train/loss" and "test/loss" are displayed consecutively without a clear grouping, making it difficult for users to distinguish different sections. The change introduces a grouping mechanism by prefix, organizing each set of prefixed metrics in a dedicated `gr.Accordion`. 

Video of `fake-training.py`:

https://github.com/user-attachments/assets/069c36ae-95a1-4a5b-b5cd-ad36ba2bf3f9

As seen in the video, up to two layers of prefixes are supported.


examples/fake-training.py
```diff
@@ -74,10 +74,12 @@ def generate_accuracy_curve(epoch, max_epochs, max_acc=0.95, min_acc=0.1):
...
         wandb.log(
             {
-                "train_loss": round(train_loss, 4),
-                "train_accuracy": round(train_accuracy, 4),
-                "val_loss": round(val_loss, 4),
-                "val_accuracy": round(val_accuracy, 4),
+                "train/loss": round(train_loss, 4),
+                "train/accuracy": round(train_accuracy, 4),
+                "train/random/seed1": random.random(),
+                "train/random/seed2": random.random(),
```
qgallouedec, line 80: It's a bit strange to log the seed, I'd suggest logging rewards, which is more aligned with what you could find with multi-objective training

```suggestion
                "train/rewards/reward1": random.random(),
                "train/rewards/reward2": random.random(),
```

abidlabs: hehe thanks I agree

qgallouedec (approved):
Screenshot 2025-08-19 at 2 19 44 PM

perfect!

PR #159, opened 2025-08-22
Provide a sync status and a call to action to upgrade the Trackio Space to have persistent storage

abidlabs:
Updates the sync status to include the last time the Trackio dashboard was synced to a dataset. 

Here's what this looks like on a live Space: https://huggingface.co/spaces/abidlabs/trackio-on-spaces-885607

https://github.com/user-attachments/assets/fa2d1178-1b11-448e-ba80-3ec91c2a37db


Saba9 (approved):
Looks good!

abidlabs:
And if persistent storage is enabled, it'll look like this:

image

abidlabs:
Thanks @Saba9!

PR #162, opened 2025-08-26
Update version to 0.3.0.dev0 and adjust publish workflow

qgallouedec:


RELEASE.md
```diff
@@ -0,0 +1,156 @@
+
```
qgallouedec, line 1:
```suggestion
```

abidlabs:
Ah it looks like PR 152 has the docs:

image

Issue #141, opened 2025-08-17
Trackio doesn't currently support logging -inf or +inf


owos:
Things that WandB handles for me that Trackio  doesn't.
1. Trackio does not handles tensor logging/
2. Trackio does  not handle instances when my logged value becomes a inf or -inf
    - Use case, logging grad_norm already starts with infinity.

These are pretty simple things ad I will keep updating the list as I find new edge cases.

PR #181, opened 2025-09-04
Add support for infinity and NaN values in logging

abidlabs:
Adds sanitization system that converts infinity values to JSON-safe strings before storage and network transmission. Deserializes them back to numeric values for proper visualization and calculations. Closes: #141


qgallouedec (approved):
lgtm!

Issue #155, opened 2025-08-21
[BUG] The column label 'train/epoch' is not unique.


lewtun:
I have created the following Space and dataset for `trackio` to log to:

* https://huggingface.co/spaces/HuggingFaceH4/Qwen3-4B-Agent
* HuggingFaceH4/Qwen3-4B-Agent_dataset

The default setting to track progress by `step` works fine, but if I switch to e.g. `train/epoch` I get an error:

```
The column label 'train/epoch' is not unique.
```

The same is also true for `train/global_step`.

<img width="1112" height="448" alt="Image" src="https://github.com/user-attachments/assets/0880ee86-5be2-4e0b-82c7-bbef9458f31e" />

abidlabs:
Thanks @hlzl for pinpointing the issue! Would you like to open a PR with the fix?

PR #182, opened 2025-09-05
fix: Select different X-axis #155

hlzl:
Fixes: #155


abidlabs (approved):
LGTM thanks for the contribution @hlzl!

PR #190, opened 2025-09-08
fix persistent storage

Saba9:
Does a few things:
- Correct our check for if persistent storage is enabled. Our previous check always returned true because `"false" is not None`
- Utilize storage when enabled. We were setting the `TRACKIO_DIR` env var to point to `/data` but the variable was never read.
- Add `space_storage` argument to `trackio.init` to allow initializing the space with storage enabled.

One caveat is that if a user initializes the space without persistent storage and then enables it prior to a dataset backup anything logged in the interim will be lost. I don't think this can be resolved. I added a warning in the upgrade message.

### DIFF
--- a/trackio/__init__.py
+++ b/trackio/__init__.py
@@ -67,7 +67,7 @@ def init(
             case the Space will be created in the currently-logged-in Hugging Face
             user's namespace. If the Space does not exist, it will be created. If the
             Space already exists, the project will be logged to it.
-        space_storage (`SpaceStorage` or `None`, *optional*, defaults to `None`):
+        space_storage ([`~huggingface_hub.SpaceStorage`] or `None`, *optional*, defaults to `None`):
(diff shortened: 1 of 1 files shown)


trackio/__init__.py
```diff
@@ -65,6 +67,8 @@ def init(
             case the Space will be created in the currently-logged-in Hugging Face
             user's namespace. If the Space does not exist, it will be created. If the
             Space already exists, the project will be logged to it.
+        space_storage ([`~huggingface_hub.SpaceStorage`] or `None`, *optional*, defaults to `None`):
+            Choice of persistent storage tier.
```
abidlabs, line 71: Perhaps its worth listing the options from the enum explicitly here in the docstring

abidlabs:
I'll go ahead and merge this in and get a release out with the fix

Issue #156, opened 2025-08-21
Wrap run names in plots


lewtun:
Currently, run names are truncated if they exceed the space on a plot (see screenshot below):

<img width="1147" height="684" alt="Image" src="https://github.com/user-attachments/assets/839fd5bf-98e0-4545-9490-796b1af0c28e" />

It would be nice if these names could be wrapped so it's easier to see which curves correspond to which run. A related point is whether we need a label for smoothed curves - perhaps it is intuitive to users from the plot alone?

Issue #129, opened 2025-08-08
In plot legend, place names `*_smoothed` at last


ghost:
If plotting a few metrics at once, important information on the color binding gets hidden, while `*_smoothed` goes first:

<img width="2534" height="810" alt="Image" src="https://github.com/user-attachments/assets/8f1032d2-ace2-48b5-8458-398d7d994868" />

A better way to sort the labels would be:
- first go "main" metric names
- then go "smoothed" metric names (if this label even needed)

Alternatively, this information could've been displayed in a few lines.

PR #193, opened 2025-09-09
Fix plot legend ordering for smoothed metrics

abidlabs:
Reorders plot legend entries so all non-smoothed runs appear first, followed by all smoothed runs. Removes the _original suffix from metric names while preserving the _smoothed suffix. Closes: #129, Closes: https://github.com/gradio-app/trackio/issues/156

Here's what it looks like now:

image


qgallouedec (approved):
Nice, I like it!

Issue #133, opened 2025-08-12
AppError: database is locked


znation:
While running locally, with 2 processes in parallel:

* jupyter notebook running my training code, which calls `wandb.log`
* trackio show hosting the dashboard

I got a random crash, which appears to come from both processes attempting to write to the SQLite DB at once. See attached traceback.

[traceback.txt](https://github.com/user-attachments/files/21733799/traceback.txt)

znation:
Good question, I am also a bit confused about what's happening here. I will dig into it a bit.

alexjwilliams:
@abidlabs I am also getting the same error. It occurs when I have my training process logging to trackio and a `trackio show` command running. This library is unusable until this is fixed since I can only run a training process for ~8k steps before logging to trackio is broken.

```
Traceback (most recent call last):
  File "/nix/store/4vdsy66j0h0j3xn9pkda8sdsvjh85cd7-python3-3.12.11-env/lib/python3.12/site-packages/gradio/queueing.py", line 667, in process_events
    response = await route_utils.call_process_api(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4vdsy66j0h0j3xn9pkda8sdsvjh85cd7-python3-3.12.11-env/lib/python3.12/site-packages/gradio/route_utils.py", line 349, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4vdsy66j0h0j3xn9pkda8sdsvjh85cd7-python3-3.12.11-env/lib/python3.12/site-packages/gradio/blocks.py", line 2274, in process_api
    result = await self.call_function(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4vdsy66j0h0j3xn9pkda8sdsvjh85cd7-python3-3.12.11-env/lib/python3.12/site-packages/gradio/blocks.py", line 1781, in call_function
    prediction = await anyio.to_thread.run_sync(  # type: ignore
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4vdsy66j0h0j3xn9pkda8sdsvjh85cd7-python3-3.12.11-env/lib/python3.12/site-packages/anyio/to_thread.py", line 56, in run_sync
[...]

PR #195, opened 2025-09-09
Fix database locking errors in concurrent access

abidlabs:
Fixes SQLite database locking errors that occur when multiple processes access the same database simultaneously. Implements a cross-process lock mechanism using file-based locking to prevent race conditions. Related: #133

Also adds a unit test to simualte aggressive race conditions. On `main`, without the lock, the test fails like this:

image


tests/test_sqlite_storage.py
```diff
@@ -60,3 +64,130 @@ def test_import_export(temp_dir):
...
+            time.sleep(0.001)
+
+    run_name = f"worker_{worker_id}"
+    operations = 0
+    db_locked_errors = 0
+
+    start_time = time.time()
+    while time.time() - start_time < duration_seconds:
+        try:
+            if random.random() < 0.6:
```
Saba9, line 103: It might be beneficial to use a fixed random seed to improve reproducibility across runs e.g. `random.seed(10)`.

Saba9 (approved):
Looks good!

tests/test_sqlite_storage.py
abidlabs: Actually just got rid of this since we only use `bulk_log` (added docstrings in `log()` to explain why it's still there)

tests/test_sqlite_storage.py
```diff
@@ -60,3 +64,130 @@ def test_import_export(temp_dir):
...
+    """
+    Worker that uses SQLiteStorage methods for database access.
+    This will be protected by ProcessLock when available.
+    """
+    import os
+    import random
+    import sqlite3
+    import time
+
+    os.environ["XDG_CACHE_HOME"] = str(temp_trackio_dir)
```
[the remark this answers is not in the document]
abidlabs: So the problem is that fixtures don't work well with multiprocessing, so I had to manually do this. Made the code a bit cleaner though.

PR #199, opened 2025-09-10
Fix warnings/errors coming from the gradio app & speed up UI

abidlabs:
Fixes a few small UI issues:

* https://github.com/gradio-app/trackio/pull/199/commits/8f1db54c42ea11a5d04fff8ae0450a1cf3a4f0c3 fixes this warning:

```
/Users/abidlabs/dev/trackio/trackio/utils.py:437: FutureWarning: DataFrameGroupBy.apply operated on the grouping columns. This behavior is deprecated, and in a future version of pandas the grouping columns will be excluded from the operation. Either pass `include_groups=False` to exclude the groupings or explicitly select the grouping columns after groupby to silence this warning.
```

* https://github.com/gradio-app/trackio/pull/199/commits/112316fb65533e2d9b0ead7f72c621584eb5dee5 skips the queue for most UI interactions. This should make the UI feel much snappier since there's no overhead of adding events to the queue.

* https://github.com/gradio-app/trackio/pull/199/commits/5b7bbeaedf53539e8e34087e025b815d7be59bde fixes this type of error that would appear in the terminal occasionally due to a race condition:

```
gradio.exceptions.Error: "Value: 'test-run-2' (type: <class 'str'>) is not in the list of choices: ['test-run-0', 'test-run-1']"
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,12 @@ jobs:
       with:
         python-version: '3.10'
         
-    - name: Install dependencies
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update -qy
+        sudo apt-get install -qy ffmpeg
+        
+    - name: Install Python dependencies
       run: |
         python -m pip install --upgrade pip
         pip install -e .[dev,tensorboard]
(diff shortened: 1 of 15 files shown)


abidlabs:
Thanks @znation!

Issue #118, opened 2025-08-06
Some info on the run in the table format (related to the config part)


abidlabs:
[internal context](https://huggingface.slack.com/archives/C08SW1X12C8/p1753991654907859?thread_ts=1753991208.424059&cid=C08SW1X12C8)

Issue #157, opened 2025-08-21
Allow runs to be deleted from the UI


lewtun:
A common workflow for me is to do a bunch of test / dummy training runs to check everything works before launching a "proper" run. As a result, I end up with lots of dummy `{username}-{timestamp}` runs in the UI which clutters things over time:

<img width="353" height="664" alt="Image" src="https://github.com/user-attachments/assets/96e63a42-d383-4561-818b-90014d1389b6" />

I'm not sure how feasible this is, but having the ability to permanently delete / hide a run in the UI would be very handy for large projects!

fcakyon:
@abidlabs this feature would be very useful for us 🤗

PR #203, opened 2025-09-10
Add a second page to show run information

abidlabs:
Adds a navbar and a second page to show some basic information for all Runs, mainly the config and timestamp created and author. Inherits all the nice sorting & filtering capabilities from the `gr.Dataframe` component. Feedback welcome on the UI:

https://github.com/user-attachments/assets/7dfa9cf9-7f02-4449-9fa2-94196c2fc920

Closes: https://github.com/gradio-app/trackio/issues/119
Closes: https://github.com/gradio-app/trackio/issues/118
Closes: https://github.com/gradio-app/trackio/issues/208

Lots of reorganization of files and setting the stage for https://github.com/gradio-app/trackio/issues/157

Quite possible that I broke something, so please do test it out! Next week, I'll work on adding more end-to-end tests to Trackio.

### DIFF
--- a/trackio/ui/fns.py
+++ b/trackio/ui/fns.py
@@ -34,7 +34,10 @@ def get_projects(request: gr.Request):
         interactive = False
     else:
         interactive = True
-        project = projects[0] if projects else None
+        if selected_project := request.query_params.get("selected_project"):
+            project = selected_project
+        else:
+            project = projects[0] if projects else None
 
     return gr.Dropdown(
         label="Project",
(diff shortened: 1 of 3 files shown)


abidlabs:
> I think it requires gradio>=4.45. Bumping the dep

Thanks!

> When the config keys don't match (quite common), it fails:

Fixed

> Also, not sure how hard it is, but, making the project persistent would make things smoother:

Agreed, will fix in a follow up PR, as it requires https://github.com/gradio-app/gradio/pulls?q=is%3Apr+is%3Aclosed to be merged.

qgallouedec (approved):
Nice! trying with sft, we've this strange rendering, because some fields are very long. Do you know if there is an easy way to enhance this? Can be done in a follow-up PR, so approved!

Screenshot 2025-09-15 at 1 37 12 PM
Screenshot 2025-09-15 at 1 37 18 PM

znation (approved):
Some nice refactoring and tests in here. Thanks! LGTM.

abidlabs:
Unfortunately, with `wrap=False`, the results are worse because the run names and other cell values are truncated:

image

So will keep as is for now, and then open an issue in Gradio to take a deeper look at what's happening.

abidlabs:
Thanks @qgallouedec and @znation for the reviews!

Issue #211, opened 2025-09-16
Increase visibility of `TRACKIO_SPACE_ID`


abidlabs:
As pointed out by @merveenoyan, the `TRACKIO_SPACE_ID` env variable is not very visible. We can:

* add some documentation for `TRACKIO_SPACE_ID` and other environment variables used by Trackio in the Trackio docs
* Provide a message like "Hint: set the TRACKIO_SPACE_ID variable to push to Space"

PR #212, opened 2025-09-17
Add documentation for environment variables

abidlabs:
Adds documentation for `TRACKIO_SPACE_ID` and other key environment variables used by Trackio. Closes: #211

cc @merveenoyan

### DIFF
--- a/docs/source/environment_variables.md
+++ b/docs/source/environment_variables.md
@@ -4,33 +4,6 @@ Trackio uses environment variables to configure various aspects of its behavior,
 
 ## Core Environment Variables
 
-### TRACKIO_SPACE_ID
-
-Sets the Hugging Face Space ID for deploying your Trackio dashboard. This is useful when you want to automatically deploy to a specific Space without passing it to `trackio.init()`. The Space will be created if it doesn't exist (requires appropriate HF token permissions).
-
-```bash
-export TRACKIO_SPACE_ID="username/space_id"
-```
-
-Or in Python:
-
-```python
-import os
-os.environ["TRACKIO_SPACE_ID"] = "username/space_id"
-```
-
-The username can be a personal Hugging Face account or an org account on Hugging Face.
-
-
-**Integration with Transformers/TRL:** When using the Transformers `Trainer` or one of the Trainer classes in TRL with `TrainingArguments(report_to="trackio")`, this environment variable will be used to determine where to deploy the dashboard. This allows you to set up Space deployment once and have all your training runs automatically use it.
-
-### TRACKIO_PROJECT_NAME
-
-Specifies the default project name for logging experiments. This is useful when you want to set a specific project name without passing it to `trackio.init()`.
-
-```bash
-export TRACKIO_PROJECT_NAME="my-project"
-```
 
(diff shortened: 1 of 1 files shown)


qgallouedec:
For TRACKIO_SPACE_ID and TRACKIO_PROJECT, I feel like we're mixing things here. These environment variable are only used with the Transformers interrogation. I could confuse reader. Plus, we already have it documented here:

https://huggingface.co/docs/trackio/transformers_integration#configuring-project-and-space

also, agree with @merve, just doing through init is actually better API-wise, I'll likely change it in transformers, so that we can only use

```python
TrainingArguments(run_name="my_run", trackio_project="my_project", trackio_space_id="my_space")
```

for the other env var, lgtm!

abidlabs:
Oh whoops I thought that these existed in Trackio and you were just referencing them in Transformers integration. I think that they should exist, no? It'll be good for consistent behavior no matter how you are using Trackio, and it'll simplify the integration on the Transformers/TRL side. I'll add these env variables to Trackio in this PR as well.

qgallouedec:
For context: the idea initially was to be aligned with the other players see 

https://github.com/huggingface/transformers/blob/58e13b9f129bb0dccc3b51e5da22f45ef3ff0ae7/src/transformers/integrations/integration_utils.py#L856-L857

https://github.com/huggingface/transformers/blob/58e13b9f129bb0dccc3b51e5da22f45ef3ff0ae7/src/transformers/integrations/integration_utils.py#L1225-L1226 (they probably aligned with WANDB in the first place)

https://github.com/huggingface/transformers/blob/58e13b9f129bb0dccc3b51e5da22f45ef3ff0ae7/src/transformers/integrations/integration_utils.py#L1906-L1907 (same here)

We **can** support these environment variables in Trackio, but I'm not sure if it should be the recommended way to use it. explicit arguments sounds like a better option to me. That's why I'm opening this PR: https://github.com/huggingface/transformers/pull/40950. Basically:

```diff
  import os
  
  import numpy as np
  from datasets import Dataset
  from transformers import Trainer, AutoModelForCausalLM, TrainingArguments
  
- os.environ["TRACKIO_PROJECT"] = "my-awesome-project"
- os.environ["TRACKIO_SPACE_ID"] = "my-awesome-username/my-awesome-space"
  
  # Create a fake dataset
  data = np.random.randint(0, 1000, (8192, 64)).tolist()
  dataset = Dataset.from_dict({"input_ids": data, "labels": data})
  
  # Train a model using the Trainer API
  trainer = Trainer(
      model=AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-0.6B"),
-     args=TrainingArguments(run_name="Qwen3-0.6B-training", report_to="trackio"),
-     args=TrainingArguments(run_name="Qwen3-0.6B-training", report_to="trackio", project="my-awesome-project", trackio_space_id="my-awesome-username/my-awesome-space"),
      train_dataset=dataset,
  )
  trainer.train()
```

abidlabs:
Ok sounds good, I'll remove these two env variables for now

abidlabs:
Made the changes, thanks @qgallouedec @merveenoyan!

PR #217, opened 2025-09-17
Allow to choose private for Trackio Space

qgallouedec:
```python
import trackio

trackio.init(
        project="my-project",
        space_id="qgallouedec/demo-private3",
        private=True,
    )
```

Screenshot 2025-09-17 at 8 20 02 AM


abidlabs (approved):
Great thanks @qgallouedec! We can consider adding an env variable for this as well.

PR #234, opened 2025-09-17
Allow deleting runs when running Trackio locally & implement Jupyter-style token authentication

abidlabs:
* Adds support for deleting runs through the Trackio UI, as requested in #157
* Note that this introduces a security risk: what if a different user accesses the Trackio dashboard (via SSH tunneling or gradio's share links), that user should NOT be able to delete runs. To address this, I've added a token-based authentication system similar to Jupyter notebooks. So now, `trackio show` generates a token on server startup and saves it through cookies and URL parameters. 
* This auth mechanism will provides write access control for future features as well.
* Note that this auth mechanism will not work on Spaces. Instead, if running on Spaces, I will replace this with a `gr.LoginButton` that can be used to oauth via Hugging Face. In that case, any user with write access to the namespace will be able to delete runs (in the same way that they can currently log runs).

Partially addresses https://github.com/gradio-app/trackio/issues/157, will close that issue once we have deleting runs supported on Spaces as well.

If you have write access, the flow looks like this (note that I'm deleting runs while other runs are being logged, so the number of runs doesn't seem to change in the first step):

https://github.com/user-attachments/assets/9c9e3e75-6c66-49f0-97aa-a8338317c0b0

If you don't, you can't click on the delete button:

https://github.com/user-attachments/assets/2c59c5fd-dedd-473f-ad68-ea35cd3776f7

I am aware that there are some UI hiccups (like the column of "true"s showing up in the dataframe). Those are bugs in Gradio that will need to be fixed upstream.

cc @lewtun for visibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
   - Visualize experiments with a Gradio dashboard locally (or, if you provide a `space_id`, on Hugging Face Spaces)
 - Everything here, including hosting on Hugging Face, is **free**!
 
-Trackio is designed to be lightweight (the core codebase is <4,000 lines of Python code), not fully-featured. It is designed in an extensible way and written entirely in Python so that developers can easily fork the repository and add functionality that they care about.
+Trackio is designed to be lightweight (the core codebase is <5,000 lines of Python code), not fully-featured. It is designed in an extensible way and written entirely in Python so that developers can easily fork the repository and add functionality that they care about.
 
 ## Installation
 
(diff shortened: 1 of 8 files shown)


qgallouedec:
Hey, just a few suggestions here: https://github.com/gradio-app/trackio/pull/238

qgallouedec (approved):
🔥

PR #238, opened 2025-09-17
Confirm deletion and avoid printing token

qgallouedec:


qgallouedec:
I was just surprised seeing a token printed. It will appear in the Space logs as well. But I've no idea how one could exploit that, maybe it's safe.

PR #240, opened 2025-09-17
QOL + align docstrings with HF standards

qgallouedec:
- Align doctoring style with rest of HF libraries, see https://github.com/huggingface/trl/issues/4036#issuecomment-3269090826
- Consistently use python 3.10+ typing style (`Optional[X]` -> `X | None`)


abidlabs (approved):
Beautiful thanks!

Issue #242, opened 2025-09-18
[Colab Notebook] Discussion of workflow


ariG23498:
As `trackio` is based on gradio and HF Spaces, we could embed the run inside a colab notebook as soon as we hit the `trackio.init` api. This would be an easier and more interactive experience for the users.

Currently we show a public url which needs to be visited by the user.

CC: @abidlabs

Issue #241, opened 2025-09-18
trackio logo does not render on public url


ariG23498:
Running trackio in colab and then visiting the *public url* does not render the trackio logo correctly (attached screenshot)

<img width="720" height="321" alt="Image" src="https://github.com/user-attachments/assets/7c8a6572-9050-48b9-b306-0e6796ad1b72" />

I have used the following on a colab notebook:
```py
import trackio
import random
import time

runs = 3
epochs = 8


for run in range(runs):
    trackio.init(
        project="my-project",
        config={"epochs": epochs, "learning_rate": 0.001, "batch_size": 64}
    )

    for epoch in range(epochs):
        train_loss = random.uniform(0.2, 1.0)
        train_acc = random.uniform(0.6, 0.95)

        val_loss = train_loss - random.uniform(0.01, 0.1)
        val_acc = train_acc + random.uniform(0.01, 0.05)

        trackio.log({
            "epoch": epoch,
            "train_loss": train_loss,
            "train_accuracy": train_acc,
            "val_loss": val_loss,
            "val_accuracy": val_acc
        })

        time.sleep(0.2)

trackio.finish()
```

PR #243, opened 2025-09-18
Improve workflow for jupyter / Colab notebooks

abidlabs:
Improves the workflow of Trackio on jupyter and Colab notebooks. See comments below for more details.

Closes: https://github.com/gradio-app/trackio/issues/241
Closes: https://github.com/gradio-app/trackio/issues/242

Confirm with this notebook: https://colab.research.google.com/drive/1KsjWZNkwOqgr5ELVGkXvXrcfpQyev6Xk?usp=sharing

Also fixes a separate issue where if a custom class was passed in as the `config`, we would preivously see an error. Now it's json-serialized in a reasonable manner.

Confirm with this notebook: https://colab.research.google.com/drive/1O0N1CMnKpG2B6J-ANvDGttOZPb0t7es0?usp=sharing which uses an object of the `LoraConfig` class as the `config`.

Running regular Python scripts should not be affected.

### DIFF
--- a/trackio/utils.py
+++ b/trackio/utils.py
@@ -729,3 +729,6 @@ def deserialize_values(metrics):
         else:
             result[key] = value
     return result
+
+def embed_url_in_notebook(url: str) -> None:
+    pass
--- a/trackio/utils.py
+++ b/trackio/utils.py
@@ -754,6 +754,8 @@ def embed_url_in_notebook(url: str) -> None:
 def to_json_safe(obj):
     if isinstance(obj, (str, int, float, bool, type(None))):
         return obj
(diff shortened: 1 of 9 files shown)


trackio/utils.py
```diff
@@ -305,12 +303,12 @@ def print_dashboard_instructions(project: str) -> None:
     Args:
         project: The name of the project to show dashboard for.
     """
-    YELLOW = "\033[93m"
+    ORANGE = "\033[38;5;208m"
```
abidlabs, line 306: This color is more visible in both light mode Colab notebooks and dark mode

abidlabs:
Thanks @znation!

PR #252, opened 2025-09-21
Add index on metrics(run_name, timestamp) for improved query performance

AkiSakurai:
Since query is ordered by timestamp.
https://github.com/gradio-app/trackio/blob/9daeadf7d1b7c89903a34b965c5c3f4b37c6f8e8/trackio/sqlite_storage.py#L374-L378


abidlabs (approved):
Thanks @AkiSakurai lgtm!

PR #253, opened 2025-09-21
use orjson for faster JSON parsing

AkiSakurai:
Add orjson dependency and update sqlite_storage to use orjson for faster JSON parsing


abidlabs (approved):
LGTM thanks @AkiSakurai! Just made a few tweaks to be explicit about orjson in the code.

PR #254, opened 2025-09-21
20250921 trackio view

mcgrof:
I work on the console mostly and so I figured I'd give a shot to get console visualization a shot. This is the result of that effort with Claude Code, and I ported over my gputop efforts onto trackio as well.


abidlabs:
This is very cool @mcgrof but is a bit out of scope and will be hard for us to maintain. Trackio is designed to be pretty lightweight and extensible so I'd recommend instead creating a separate trackio-view pypi package which we could link to from our documentation. Let us know if that would be of interest to you.

Issue #124, opened 2025-08-07
Add run groups


lewtun:
A common workflow when running many experiments is to group them by domain / algorithm etc, e.g. in post-training one typically does SFT followed by DPO / RL.

It would be useful is `trackio` allowed one define a "run group" under which all runs can be gathered (e.g. all SFT runs are grouped under an ``model-name-sft` and ditto for DPO with `model-name-dpo`)

WandB has this functionality and it's very helpful when trying to handle 100s of runs: https://docs.wandb.ai/guides/runs/grouping/

Note: if supported in `trackio`, we should also expose it as an env var in `transformers`: https://github.com/huggingface/transformers/blob/555cbf59178134d1b713a7022129aaddfe6e70cf/src/transformers/integrations/integration_utils.py#L1094

lewtun:
Yes that's right - see the example below from WandB. Basically I have a project called `open-r1` and within that project are various groups like `r1-distill-qwen-3b`. Being able to toggle on / off which groups are displayed is very helpful when a single project may involve many different types of experiments

<img width="407" height="412" alt="Image" src="https://github.com/user-attachments/assets/5eeff76b-d7ed-4521-b428-56f495bbf92e" />

nicolas-dufour:
Similarly, would be cool to have tags, so we can easily filter runs

PR #257, opened 2025-09-23
allow grouping runs in the sidebar by either their group or config parameters

Saba9:
Allows specifying a run's `group` during initialization and updates the UI to add a new "Group by..." dropdown which if used will group runs by whatever field was selected. Available fields are those included in the run's config (excluding `_Created`).

trackio-video-groups-2

(Closes: #124)

### DIFF
--- a/trackio/ui/run_detail.py
+++ b/trackio/ui/run_detail.py
@@ -14,6 +14,7 @@ except ImportError:
 RUN_DETAILS_TEMPLATE = """
 ## Run Details
 * **Run Name:** `{run_name}`
+* **Group:** `{group}`
 * **Created:** {created} by {username}
 """
 
@@ -47,6 +48,8 @@ with gr.Blocks() as run_detail_page:
         if not config:
             return gr.Markdown("No run details available"), {}
 
+        group = config.get("_Group", "None")
+
         created = config.get("_Created", "Unknown")
         if created != "Unknown":
             created = utils.format_timestamp(created)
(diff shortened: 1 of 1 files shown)


trackio/ui/main.py
```diff
@@ -25,6 +26,7 @@
     from trackio.ui.run_detail import run_detail_page
     from trackio.ui.runs import run_page
 except ImportError:
+    import ui.helpers.run_selection as rs
```
abidlabs, line 29:
```suggestion
    from ui.helpers import run_selection
```

trackio/ui/fns.py
```diff
@@ -56,3 +63,41 @@ def update_navbar_value(project_dd):
...
+    config_key = CONFIG_COLUMN_MAPPINGS_REVERSE.get(config_key, config_key)
+    configs = SQLiteStorage.get_all_run_configs(project)
+    groups: dict[str, list[str]] = {}
+    for run_name, config in configs.items():
+        if filter_text and filter_text not in run_name:
+            continue
+        group_name = config.get(config_key, "None")
+        label = f"{display_key}: {group_name}"
+        groups.setdefault(label, []).append(run_name)
+    # sort within each group
```
abidlabs, line 98:
```suggestion
```

abidlabs (approved):
Beautiful PR @Saba9! Tested and works great. Left a few small stylistic suggestions.

The only other thing is that it would be great if we could list the group name in the Run Detail page if the run has a group. In `trackio/ui/run_detail.py`

trackio/ui/helpers/run_selection.py
```diff
@@ -0,0 +1,96 @@
+from dataclasses import dataclass, field
+
+import gradio as gr
+
+
+@dataclass
+class RunSelection:
+    choices: list[str] = field(default_factory=list)
+    selected: list[str] = field(default_factory=list)
+    locked: bool = False
```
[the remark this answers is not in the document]
Saba9: It serves the same purpose as `user_interacted_with_runs` did prior to this PR. It's set to true after user interaction so that we don't override the user's run selections with those in the URL.

Issue #258, opened 2025-09-23
Suppress HTTP requests from logs by default


lewtun:
Currently I see a lot of HTTP requests in my training logs:

```
[default0]:2025-09-23 17:17:49 - INFO - httpx - HTTP Request: GET https://huggingfaceh4-qwen3-4b-agent.hf.space/config "HTTP/1.1 200 OK"
[default0]:2025-09-23 17:17:50 - INFO - httpx - HTTP Request: GET https://huggingfaceh4-qwen3-4b-agent.hf.space/gradio_api/info?serialize=False "HTTP/1.1 200 OK"
[default0]:2025-09-23 17:17:50 - INFO - httpx - HTTP Request: GET https://huggingfaceh4-qwen3-4b-agent.hf.space/gradio_api/heartbeat/9ed809bb-50ed-4cac-8efb-32befa1e95d3 "HTTP/1.1 200 OK"
```

It would be nice if these could be suppressed by default as they clutter / explode the logs quite a bit.

PR #266, opened 2025-10-01
Suppress httpx INFO logs by default

abidlabs:
Adds logging configuration to suppress httpx INFO level logs by setting the httpx logger to WARNING level. Prevents HTTP request logs from cluttering training output while still allowing warnings and errors to appear. Test with a script like this:

```py
import logging
import sys

logging.basicConfig(
    level=logging.INFO, format="%(levelname)-8s %(name)-15s %(message)s"
)

import trackio

try:
    from gradio_client import Client

    # This will make HTTP requests that normally produce httpx INFO logs
    print("\nAttempting connection to a Hugging Face Space...\n")
    try:
        client = Client("abidlabs/en2fr", verbose=False)
        print("\n✓ Connection attempt completed")
    except Exception as e:
        print(
            f"\n✓ Connection failed (expected): {e[:100] if len(str(e)) > 100 else e}"
        )

except ImportError:
    print("gradio_client not available")
```

Closes: https://github.com/gradio-app/trackio/issues/258

### DIFF
--- a/trackio/__init__.py
+++ b/trackio/__init__.py
@@ -12,8 +12,6 @@ from gradio.themes import ThemeClass
 from gradio_client import Client
 from huggingface_hub import SpaceStorage
 
-logging.getLogger("httpx").setLevel(logging.WARNING)
-
 from trackio import context_vars, deploy, utils
 from trackio.imports import import_csv, import_tf_events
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
👌

Issue #263, opened 2025-09-27
Can't delete run in private space


pnguyen-dh:
Hi, 

firstly, thank you for your working, and this make my works 100% easier, so I'm very appreciating your work.

And as the title suggest, I can't delete my run in space. So I set the space_id correct, I can see trackio in space, but it shows  "Need write access to delete runs."

Do I mess up anything in the setup part? Or for now, I can't delete the run in space? Thank you

Best,
Phuong

abidlabs:
This is a bug on our part. Deleting runs works locally but not on Spaces yet, will fix this week!

PR #268, opened 2025-10-01
Allow deleting runs on Spaces (via Oauth) and on Colab

abidlabs:
Adds a `gr.LoginButton` to allow deleting runs from the UI on Spaces. Try by running `python examples/deploy-on-spaces.py` and then clicking the login button. You should then be able to delete runs:

https://github.com/user-attachments/assets/e6b000c4-d49e-4ef5-ba60-2dfe5eda6730

Closes: https://github.com/gradio-app/trackio/issues/263
Closes: https://github.com/gradio-app/trackio/issues/157

Fixes a separate bug that was preventing deleting runs on Colab/Jupyter notebooks (see below)


trackio/ui/fns.py
```diff
@@ -49,10 +53,102 @@ def get_projects(request: gr.Request):
...
+                    "Expected the provided hf_token with fine grained permissions to provide write access to the space"
+                )
+        # reject read-only tokens
+        elif access_token["role"] != "write":
+            raise PermissionError(
+                "Expected the provided hf_token to provide write permissions"
+            )
+
+
+def check_oauth_token_has_write_access(oauth_token: str | None) -> None:
```
abidlabs, line 131: The result from running whoami on an oauth token has a different format than when whoami is run on a regular hf token, hence the separate method.

trackio/ui/main.py
```diff
@@ -625,12 +572,17 @@ def create_media_section(media_by_run: dict[str, dict[str, list[MediaData]]]):
...
+        // Only remove write_token from URL if not in iframe
+        // In iframes, keep it in URL as cookies may be blocked
+        const inIframe = window.self !== window.top;
+        if (!inIframe) {
+            urlParams.delete('write_token');
+            const newUrl = window.location.pathname + 
+                (urlParams.toString() ? '?' + urlParams.toString() : '') + 
+                window.location.hash;
+            window.history.replaceState({}, document.title, newUrl);
+        }
```
abidlabs, line 585: This is the fix for colab/jupyter notebooks or other embedded environments.

PR #317, opened 2025-10-15
quick fixes for trackio.histogram

abidlabs:
Some fixes needed to deploy histograms successully on Spaces: https://huggingface.co/spaces/abidlabs/histogram-test-208105

will merge this in so that we can do the release


abidlabs:
Will merge this in so that we can get the release out.

PR #341, opened 2025-11-07
Adds a basic UI test to `trackio`

abidlabs:
In `tests/ui/test_ui_display.py`, we add a UI test using `playwright`. This is a proof-of-concept that we'll expand to cover other cases as well (although we don't want to add too many playwright tests since they are time-consuming to run -- just this one takes 13s).


abidlabs:
Going to merge this in so that we can start adding more tests. Happy to revert anything if anyone has feedback on this.

Issue #303, opened 2025-10-13
offline tracking, online syncing


candemircan:
Hi there, 

Thanks for this cool project.  I was wondering whether it'd be possible to track runs in a python script in an offline manner, and sync them to an HF space through the CLI separately. 

Following the wandb API, this could look something like

```python
trackio.init(project="my-project", space_id="orgname/space_id", mode="offline")
```

and in the shell you can run

```bash
trackio sync "my-project" --include-offline
```

Many academic clusters have internet only on the login node and not on the compute nodes. This type of decoupling would be very helpful for such workflows.

If you would consider adding this feature to your roadmap, I'd be happy to help :)

Many thanks!
Can

abidlabs:
Hi @candemircan do you have bandwidth to take this on?

candemircan:
Hi @abidlabs, apologies for the delay here! 

Yes, i will implement this over the weekend and make a pull request. Hope that's a good timeline for you still. Let me know if not, and I can try to get around to it earlier

abidlabs:
Sounds good thanks @candemircan! We're planning on using this quite a bit internally at Hugging Face, so if you have time to put togehter a PR this weekend, that'd be great.

In terms of an API, how about we do:

```
trackio push --project "project_name" --space_id "orgname/space_id"
```

for the CLI, and

```
trackio.push(project_name="project_name", space_id="orgname/space_id")
```

for the Python

Thanks!

PR #343, opened 2025-11-08
Sync offline projects to Hugging Face spaces

candemircan:
## Short description

This PR adds `trackio.push` and a corresponding CLI as discussed in #303 . It allows users to push an existing local project to a specified Hugging Face Space or to sync the local changes to the Space. Corresponding docs and tests are also added.

The new functionality can be used as follows:

```sh
trackio push --project "my-project" --space-id "username/space_id"
```

```py
trackio.push(project="my-project", space_id="username/space_id")
```

> [!WARNING]
> I removed the snippet from the `upload_db_to_space` function on the `ui` side that checks whether `db_project_path` exists and raises an error if so. 
> Without doing this, it was not possible to do something like:
> ```py
> import trackio
>
> # fully new project
> run = trackio.init(project="my-project", name="model-1")
> run.log({"accuracy": 0.98}, step=1)
> run.finish()
> 
> # this works well
> trackio.push(project="my-project", space_id="username/space_id") 
>
> run = trackio.init(project="my-project", name="model-2")
> run.log({"accuracy": 0.90}, step=1)
> run.finish()
>
> # this raises a database already exists error
> trackio.push(project="my-project", space_id="username/space_id") 
> ```
> If removing this check has unintended consequences, I can just add that line again. However, if I understand the repo correctly, then we would need more major changes to sync different runs from the same project at different times.

## AI Disclosure

-----

- [x] I used AI to familiarize myself with the codebase and understand the general design, though wrote the code myself
- [ ] I did not use AI

----

## Type of Change

- [ ] Bug fix
- [x] Documentation update
- [x] Test improvements

## Related Issues

Closes: #303 

## Testing and linting

Done.

### DIFF
--- a/trackio/package.json
+++ b/trackio/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "trackio",
-	"version": "0.8.2",
+	"version": "0.8.1",
 	"description": "",
 	"python": "true"
 }


trackio/ui/main.py
```diff
@@ -293,10 +293,6 @@ def upload_db_to_space(
     """
     fns.check_hf_token_has_write_access(hf_token)
     db_project_path = SQLiteStorage.get_project_db_path(project)
-    if os.path.exists(db_project_path):
-        raise gr.Error(
-            f"Trackio database file already exists for project {project}, cannot overwrite."
-        )
```
abidlabs: We can remove this, but I think we should instead have a check in `deploy.upload_db_to_space` to see if the project already exists and if so, ask the user for confirmation as this would overwrite any existing data in the trackio project.

abidlabs (commented):
Thanks so much for this PR @candemircan! Generally looks good to me, I'll do some more testing, but have just one suggestion above

abidlabs (approved):
Made some tweaks but overall, great PR, tysm @candemircan!

Issue #363, opened 2025-12-15
Feature request: 'latest run' view.


davidgilbertson:
It would be nice to have a URL and/or tickbox in the dashboard so that the charts I see are always for the latest run. Right now, the typical workflow most of the time is:
1. Kick off a new run locally
2. Go to the dashbaord
3. Wait for the new run to show up
4. Untick the older runs I don't want to see any more

An 'always show the latest run' view would see a new view and change the UI so that only that run shows.

Having a view that's _always_ the latest run would really make this more dashboardy, not something I need to actually interact with.

abidlabs:
Closed via: https://github.com/gradio-app/trackio/pull/367

PR #367, opened 2025-12-18
Add option to only show latest run, and fix the double logo issue.

abidlabs:
Some changes to the underlying Gradio app that:

- [x] Fixes double logo issue: see https://huggingface.co/spaces/abidlabs/trackio-on-spaces-579957

- [x] Closes: https://github.com/gradio-app/trackio/issues/363

image


abidlabs:
I'll go ahead and merge this in after CI passes, since it's only UI changes and so that we can get it out for the release!

PR #374, opened 2025-12-29
Save Optimized Parquet files

lhoestq:
Updated to save DataFrame as an optimized Parquet file with page index and content defined chunking.

cc @kszucs for viz

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
         
     - name: Install Python dependencies
       run: |
-        uv pip install --system -e .[dev,tensorboard] --prerelease=allow
+        uv pip install --system -e .[dev,tensorboard,spaces] --prerelease=allow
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -13,6 +13,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     "pandas<3.0.0",
+    "pyarrow>=21.0.0",  # for parquet content-defined chunking
```
abidlabs: We don't need to add this dependency to the core trackio library, as `export_to_parquet()` is only called from Spaces, and we already automatically install `pyarrow` when Trackio is running on Spaces here: https://github.com/gradio-app/trackio/blob/137e0eaf52da21993ec2d693f6c1777897dbd61e/trackio/deploy.py#L103

That being said, we can probably clean this logic up a bit.

PR #382, opened 2026-01-07
Fix app file path when Trackio is not installed from source

abidlabs:
Internal context: https://huggingface.slack.com/archives/C08SW1X12C8/p1767781217949789


abidlabs:
Going to merge so that we can do a release!

PR #411, opened 2026-01-28
Fix test hang caused by infinite retry loop in background thread

znation:
The background client thread in Run._init_client_background() was stuck in an infinite Fibonacci backoff retry loop when connection to the Gradio server failed. This caused tests to hang indefinitely at run.finish() when trying to join the thread.

Fixed by:
- Adding check for _stop_flag at the start of each retry iteration
- Using _stop_flag.wait() instead of time.sleep() during backoff to allow immediate interruption when finish() is called

This ensures background threads can be properly terminated during test cleanup, preventing indefinite hangs.

## AI Disclosure

We encourage the use of AI tooling in creating PRs, but the any non-trivial use of AI needs be disclosed. E.g. if you used Claude to write a first draft, you should mention that. Trivial tab-completion doesn't need to be disclosed. **You should self-review all PRs, especially if they were generated with AI**. 

-----

- [x] I used AI (Claude Code w/ Sonnet) to make all changes and write the commit message.
- [ ] I did not use AI

----

## Type of Change

- [x] Bug fix
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

n/a

## Testing and linting

   ```bash
   python -m pytest
   ```

   ```bash
   ```

### DIFF
--- a/.changeset/heavy-lights-end.md
+++ b/.changeset/heavy-lights-end.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Check `dist-packages` when checking for source installation
--- a/trackio/deploy.py
+++ b/trackio/deploy.py
@@ -43,7 +43,7 @@ def _is_trackio_installed_from_source() -> bool:
     """Check if trackio is installed from source/editable install vs PyPI."""
     try:
         trackio_file = trackio.__file__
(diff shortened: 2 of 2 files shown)


znation:
This no longer seems necessary as I am unable to reproduce the hang. Closing without merging.

PR #420, opened 2026-02-04
fix URL of images in media gallery

catwell:
## Short description

This PR fixes a bug where in some cases, in particular when using a SSH tunnel, URLs of images are incorrect and include `gradio_api` twice, e.g. `http://127.0.0.1:7861/gradio_api/run/predict/gradio_api/file=...`. A workaround is setting `GRADIO_ROOT_PATH` but it must be set e.g. to `http://127.0.0.1:7861/` (the empty string does not work, `/` works for this component but breaks the rest of the app).

To avoid this we construct the HTTP URL instead of passing the file path when `GRADIO_ROOT_PATH` is unset.

## AI Disclosure

I used AI (Claude Code and Gemini) to investigate the problem, but I wrote the patch without assistance.

----

## Type of Change

- [x] Bug fix
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

None I know about.

## Testing and linting

Code is linted / formatted.

I could not run all the tests because of dependency problems but those I could run pass.

### DIFF
--- a/trackio/ui/media_page.py
+++ b/trackio/ui/media_page.py
@@ -1,5 +1,6 @@
 """The Media and Tables page for the Trackio UI."""
 
+import os
 import re
 from dataclasses import dataclass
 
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks for the PR @catwell. Could you provide me with a way to repro the original issue? Sounds like a bug underlying in Gradio. Either way, we can fix it, but we'll need a repro to verify the issue and fix.

catwell:
I've given Claude my repro and the Gradio source and it thinks it found the bug :D

image

catwell:
@abidlabs Could you confirm the issue? I do not know Gradio well enough to know 100% if the patch suggested by Claude is the right one but if I apply it locally it fixes the issue indeed.

Issue #332, opened 2025-11-05
When embedding trackio plots, have a way to embed just the charts without the accordion


abidlabs:
As suggested by @lewtun: 

> A small suggestion on my side would be to have a way to embed just the plot, without the charts (1) stuff on the outside

PR #435, opened 2026-02-21
feat: allow hiding section header accordions

Saba9:
## Short description

This PR adds a "Show section headers" checkbox below the embed options panel (default checked). When unchecked, accordion wrappers around metric groups are hidden. The state is also encoded in the iframe URL so that the embed reflects the setting.

### Shown (default)

image

### Hidden

image

## AI Disclosure

- [x] I used AI to implement the feature and tweaked it myself

## Type of Change

## Related Issues

Closes: #332

## Testing and linting

- [x] `python -m pytest` — 89 passed, 11 skipped
- [x] `ruff check --fix --select I && ruff format` — all checks passed

### DIFF
--- a/trackio/ui/main.py
+++ b/trackio/ui/main.py
@@ -752,10 +752,6 @@ with gr.Blocks(title="Trackio Dashboard") as demo:
             language="html",
             visible=bool(os.environ.get("SPACE_HOST")),
         )
-        show_headers_cb = gr.Checkbox(
-            label="Show section headers",
-            value=True,
-        )
         with gr.Group():
             run_tb = gr.Textbox(label="Runs", placeholder="Type to filter...")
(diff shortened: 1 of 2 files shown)


znation (approved):

trackio/ui/main.py
```diff
@@ -680,6 +702,11 @@ def configure(request: gr.Request):
             language="html",
             visible=bool(os.environ.get("SPACE_HOST")),
         )
+        show_headers_cb = gr.Checkbox(
+            label="Show section headers",
+            value=True,
+            visible=bool(os.environ.get("SPACE_HOST")),
```
abidlabs: Why do we only show this checkbox on Spaces? This setting option is valuable locally as well

Saba9: Ah, I'll fix that.

PR #450, opened 2026-03-11
Use Svelte source for Gradio components directly in Trackio dashboard

abidlabs:
Although this increases the codebase significantly (primarily because we are vendoring in the Svelte code for the Gradio components), this PR has several advantages which make it worth it imo:

* Major performance improvements since we no longer need to make backend requests for every UI interaction.
* This gives us complete control over the UI / UX of the Trackio dashboard
* This opens the door to having a static (read-only) version of the Trackio dashboard

We still keep the Gradio backend server because it manages API & MCP access.

You can see this deployed in a Space here: https://huggingface.co/spaces/abidlabs/trackio-on-spaces-545047

https://github.com/user-attachments/assets/bab93689-3e90-436e-93d1-a4ac71b9647f


trackio/sqlite_storage.py
```diff
@@ -844,7 +844,25 @@ def has_system_metrics(project: str) -> bool:
...
+                    (run,),
+                )
+                return cursor.fetchone()[0]
+        except sqlite3.OperationalError as e:
+            if "no such table: metrics" in str(e):
+                return 0
+            raise
+
+    @staticmethod
+    def get_logs(project: str, run: str, max_points: int = 1500) -> list[dict]:
```
znation: Any callers of `get_lots(project, run)` will now get 1500 max_points whereas previously there was no default max. This could be considered an API breaking change -- do we expect there to be any consumers of this function that haven't been updated within this PR, or are external to this repo, that might expect the previous behavior?

PR #454, opened 2026-03-26
Bar chart single point

abidlabs:
Renders single-value metrics as bar chart instead of line plots:

https://github.com/user-attachments/assets/3515d758-6ff6-4117-8dc1-f25e7c3c86eb


qgallouedec (approved):
Looks good! I cannot really review the svelte code though 😅

PR #481, opened 2026-04-09
Add multi-GPU system metrics support

Saba9:
## Summary
- **Backend**: `GpuMonitor` now queries all physical GPUs from any process by ignoring `CUDA_VISIBLE_DEVICES`, so rank 0 can collect metrics for every GPU on the machine during distributed training (uses pynvml's `nvmlDeviceGetCount()` directly)
- **Frontend**: System Metrics page renders per-GPU sub-accordions (collapsed by default) when multiple GPUs are detected, showing utilization, allocated memory, power, and temperature per GPU
- **Single-GPU**: UI is unchanged — no sub-accordions, same summary metrics as before
- **Manual API**: `trackio.log_gpu()` still respects `CUDA_VISIBLE_DEVICES`

## Changes
- `trackio/gpu.py` — Add `get_all_gpu_count()`, add `all_gpus` param to `collect_gpu_metrics()`, update `GpuMonitor` to use them
- `trackio/frontend/src/pages/SystemMetrics.svelte` — Add subgroup rendering for multi-GPU, strip `gpu/` prefix from summary chart titles
- `tests/unit/test_gpu.py` — Unit tests for `get_all_gpu_count()` and `collect_gpu_metrics(all_gpus=True/False)`
- `tests/e2e-local/test_basic_logging.py` — Update existing mock, add multi-GPU e2e test
- `examples/test_multi_gpu_mock.py` — Mock script to test 4-GPU UI locally
- `examples/test_single_gpu_mock.py` — Mock script to test single-GPU UI locally

## Test plan
- [x] `pytest tests/unit/test_gpu.py` — 6 tests pass
- [x] `pytest tests/e2e-local/test_basic_logging.py` — 7 tests pass (including new multi-GPU test)
- [x] `pytest` — full suite passes (1 pre-existing flaky failure in `test_import_export`)
- [x] Frontend builds cleanly
- [x] Manual test with `examples/test_multi_gpu_mock.py` → verify System Metrics shows per-GPU accordions
- [x] Manual test with `examples/test_single_gpu_mock.py` → verify single-GPU UI unchanged
- [x] Test on actual multi-GPU machine


abidlabs:
Thanks @Saba9 were you able to test on a multiGPU machine (potentially with HF jobs)? Would be great to see how it looks

Saba9:
@abidlabs Not yet. I ran tests where I replaced `pynvml` with `MagicMock` to simulate multi-gpu. I'll try running it with HF jobs and post pictures soon.

Saba9:
@abidlabs Tested it with HF jobs on a dual GPU machine. Seems to be working!

**Default View**
Screenshot 2026-04-09 at 9 44 54 AM
**Per-GPU metrics expanded**
Screenshot 2026-04-09 at 9 46 22 AM

tests/e2e-local/test_basic_logging.py
```diff
@@ -156,3 +156,49 @@ def fake_gpu_metrics(device=None):
     assert log["gpu/0/power"] == 150.0
     assert log["gpu/mean_utilization"] == 75
     assert "timestamp" in log
+
+
+def test_auto_log_gpu_multi(temp_dir):
```
abidlabs, line 161: Not sure if this test is adding much since everything is mocked.

tests/unit/test_gpu.py
```diff
@@ -25,3 +26,73 @@ def test_reset_energy_baseline():
     gpu._energy_baseline = {0: 1000.0, 1: 2000.0}
     gpu.reset_energy_baseline()
     assert gpu._energy_baseline == {}
+
+
+def _make_mock_pynvml(num_gpus=4):
```
abidlabs, line 31: Again, I don't think we really need to create this whole mock fixture to test whether the gpus are being counted correctly? I think it'd be better to remove or replace with a simpler test

abidlabs:
Amazing, @Saba9! I was exploring the UI, and I think it might be useful to actually plot the the system metrics from multiple GPUs on the same graph, as users may want to compare metrics across the different GPUs easily? What do you think -- here's how wandb seems to do it for reference:

image

I know this might get a bit crowded but what we could do is, for the System Metrics page, have a list of devices/gpus in the left sidebar, just like we have runs, allowing people to trim the number of devices if it becomes too unwieldy

cc @qgallouedec @kashif for visibility

kashif:
thanks! looks good

abidlabs:
Made some UI tweaks to put all of the devices on the same graph @Saba9! 

image

https://huggingface.co/spaces/abidlabs/pr-481-multigpu-demo-20260415-1935

Everything else LGTM, so I'll go ahead and merge this in after CI is green

PR #482, opened 2026-04-09
Use server-side bucket copy when freezing Spaces

abidlabs:
- require `huggingface_hub>=1.10.0` and use `copy_files()` for bucket-to-bucket media copies during `freeze()`
- keep `freeze()` documented as a point-in-time snapshot rather than a live mirror

cc @Wauplin for visibility and appreciation for the new `copy_files()` method!

### DIFF
--- a/docs/source/deploy_embed.md
+++ b/docs/source/deploy_embed.md
@@ -60,7 +60,7 @@ trackio freeze --space-id "username/my-space" --project "my-project"
 
 This creates a new static Space (by default named `{space_id}_static`) containing a snapshot of the project's data from the source Space's bucket. The original Space is not modified.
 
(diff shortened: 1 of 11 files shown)


docs/source/deploy_embed.md
```diff
@@ -60,6 +60,8 @@ trackio freeze --space-id "username/my-space" --project "my-project"
 
 This creates a new static Space (by default named `{space_id}_static`) containing a snapshot of the project's data from the source Space's bucket. The original Space is not modified.
 
+Unlike `trackio.sync(..., sdk="static")`, `freeze()` is a one-time snapshot. If new metrics are later uploaded to the original Gradio Space, the frozen static Space will not update automatically.
```
abidlabs:
```suggestion
Note that`freeze()` is a one-time snapshot. If new metrics are later uploaded to the original Gradio Space, the frozen static Space will not update automatically.
```

Wauplin (approved):
Glad it already proves useful! 😀

PR #484, opened 2026-04-10
Fix duplicate columns in parquet export

abidlabs:
use `pd.concat(..., axis=1)` to avoid DataFrame fragmentation in the flattening path, avoids this warning in the terminal:

```
pandas PerformanceWarning: the DataFrame was highly fragmented...
```


znation (commented):
Looks good, thanks for fixing!

PR #489, opened 2026-04-10
Remove `gradio` dependency in `trackio` -- only `gradio_client` is needed locally anymore. Also lazily import `pandas` and remove it as a dependency

abidlabs:
This PR drops the dependency on `gradio`, `fastapi`, and `pandas` for local Trackio usage. The dashboard and CLI now rely on `gradio-client`, Starlette, and Uvicorn, with small vendored copies of the Gradio pieces we still need (tunneling, networking helpers, and shared exceptions). Server startup and ASGI wiring were refactored accordingly (`launch.py`, `asgi_app.py`, `server.py`, etc.). Note that when we deploy on Gradio Spaces, we still use `gradio` as but not locally.

Before: Non-editable install (`pip install .`) on macOS, Python 3.10.19. Figures are total size of `site-packages` after dependency resolution.

| | Before (`main`) | After (this branch) |
| --- | ---: | ---: |
| `site-packages` on disk | ~230.8 MB | ~99.9 MB |
| Packages (`pip freeze`) | 57 | 37 |

Comparing to other packages:

| Package | `site-packages` on disk | Packages (`pip freeze`) |
| --- | ---: | ---: |
| Trackio (this branch) | ~99.9 MB | 37 |
| wandb | ~119.6 MB | 20 |
| mlflow | ~558.5 MB | 91 |

(It feels like the local Trackio dashboard is even snappier now though that's just based on my vibes)

### DIFF
--- a/trackio/_vendor/networking.py
+++ b/trackio/_vendor/networking.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+import os
+import time
+import warnings
+from pathlib import Path
+from urllib.parse import urlparse, urlunparse
+
+import httpx
+
+from trackio._vendor.gradio_exceptions import ShareCertificateWriteError
+from trackio._vendor.tunneling import CERTIFICATE_PATH, Tunnel
+
+GRADIO_API_SERVER = "https://api.gradio.app/v3/tunnel-request"
+GRADIO_SHARE_SERVER_ADDRESS = os.getenv("GRADIO_SHARE_SERVER_ADDRESS")
+
+
+def setup_tunnel(
+    local_host: str,
+    local_port: int,
+    share_token: str,
+    share_server_address: str | None,
+    share_server_tls_certificate: str | None,
+) -> str:
+    share_server_address = (
+        GRADIO_SHARE_SERVER_ADDRESS
+        if share_server_address is None
+        else share_server_address
+    )
+    if share_server_address is None:
+        try:
+            response = httpx.get(GRADIO_API_SERVER, timeout=30)
+            payload = response.json()[0]
+            remote_host, remote_port = payload["host"], int(payload["port"])
+            certificate = payload["root_ca"]
+        except Exception as e:
+            raise RuntimeError(
+                "Could not get share link from Gradio API Server."
+            ) from e
+        try:
+            Path(CERTIFICATE_PATH).parent.mkdir(parents=True, exist_ok=True)
+            with open(CERTIFICATE_PATH, "w") as f:
+                f.write(certificate)
+        except Exception as e:
+            raise ShareCertificateWriteError(
+                f"{e}. This can happen if the current working directory is read-only."
+            ) from e
+        share_server_tls_certificate = CERTIFICATE_PATH
+
+    else:
+        remote_host, remote_port = share_server_address.split(":")
+        remote_port = int(remote_port)
+    tunnel = Tunnel(
+        remote_host,
+        remote_port,
+        local_host,
+        local_port,
+        share_token,
+        share_server_tls_certificate,
+    )
+    address = tunnel.start_tunnel()
+    return address
+
+
+def url_ok(url: str) -> bool:
+    try:
+        for _ in range(5):
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore")
+                r = httpx.head(url, timeout=3, verify=False)
+            if r.status_code in (200, 401, 302, 303, 307):
+                return True
+            time.sleep(0.500)
+    except (ConnectionError, httpx.ConnectError, httpx.TimeoutException):
+        return False
+    return False
+
+
+def normalize_share_url(share_url: str, share_server_protocol: str) -> str:
+    parsed_url = urlparse(share_url)
+    return urlunparse((share_server_protocol,) + parsed_url[1:])
--- a/trackio/asgi_app.py
+++ b/trackio/asgi_app.py
@@ -1,175 +1,167 @@
 from __future__ import annotations
 
-import asyncio
 import inspect
 import json
-import secrets
-import traceback
-from collections.abc import Callable
+import math
+import tempfile
+from pathlib import Path
 from typing import Any
+from urllib.parse import unquote
 
 from starlette.applications import Starlette
 from starlette.requests import Request
-from starlette.responses import FileResponse, JSONResponse, Response, StreamingResponse
+from starlette.responses import FileResponse, JSONResponse, Response
 from starlette.routing import Route
 
 from trackio.exceptions import TrackioAPIError
+from trackio.remote_client import HTTP_API_VERSION
 
-API_PREFIX = "/gradio_api"
 
-
-def _serialize_result(data: Any) -> str:
-    def default(o: Any):
-        if isinstance(o, (dict, list, str, int, float, bool)) or o is None:
-            return o
-        if hasattr(o, "item"):
-            try:
-                return o.item()
-            except Exception:
-                pass
-        return str(o)
-
-    return json.dumps(data, default=default)
+def _json_safe(data: Any) -> Any:
+    if data is None or isinstance(data, (str, bool, int)):
+        return data
+    if isinstance(data, float):
+        return data if math.isfinite(data) else None
+    if isinstance(data, dict):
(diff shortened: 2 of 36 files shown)


abidlabs:
@copilot do another pass over the latest changes

abidlabs:
@copilot review this PR again

trackio/asgi_app.py
```diff
@@ -0,0 +1,231 @@
...
+    elif body is not None:
+        args = [body]
+
+    if not isinstance(args, list):
+        args = [args]
+    if not isinstance(kwargs, dict):
+        kwargs = {}
+
+    try:
+        result = _invoke_handler(fn, request, args=args, kwargs=kwargs)
```
znation, line 172: Claude thinks this should be parallelized with asyncio.get_event_loop().run_in_executor(None, ...) or switch to anyio.to_thread.run_sync; I'm personally not so sure since this will ultimately result in DB operations that may need to be serialized anyway (since SQLite is not so parallel w/ writes).

trackio/mcp_setup.py
```diff
@@ -0,0 +1,155 @@
...
+    import trackio.server as trackio_server  # noqa: PLC0415
+
+    if os.getenv("SYSTEM") == "spaces":
+        try:
+            trackio_server.check_hf_token_has_write_access(hf_token)
+        except PermissionError as e:
+            raise ValueError(str(e)) from e
+        return
+
+    if write_token != trackio_server.write_token:
```
znation: String != comparison is not constant-time. Use `not secrets.compare_digest(write_token or "", trackio_server.write_token)` to prevent timing-oracle attacks on
  the write token.

  Same issue: server.py:335 parts[1] == token.

trackio/remote_client.py
```diff
@@ -1,28 +1,179 @@
...
+class RemoteClient:
+    def __init__(
+        self,
+        space: str,
+        hf_token: str | None = None,
+        httpx_kwargs: dict[str, Any] | None = None,
+        verbose: bool = False,
+    ) -> None:
+        self._space = space
+        try:
```
znation, line 163: ValueError from invalid src format gets misleadingly wrapped — remote_client.py:162-176
```
  try:
      if _supports_http_api(space, ...):
          ...
      else:
          ...
  except Exception as e:
      raise ConnectionError(f"Could not connect to Space '{space}'. Is it running?\n{e}") from e
```
  If space is an invalid format (e.g., a bare repo name without a /), _resolve_src_url raises ValueError — but it gets caught here and re-raised as
  ConnectionError("... Is it running?"), which is a misleading message for what is actually a bad argument. The ValueError should be re-raised directly, not
  wrapped.

trackio/asgi_app.py
```diff
@@ -0,0 +1,231 @@
...
+    return JSONResponse({"paths": saved_paths})
+
+
+async def file_handler(request: Request) -> Response:
+    fs_path = request.query_params.get("path")
+    if fs_path is None:
+        return Response("Missing path", status_code=400)
+    fp = Path(unquote(fs_path))
+    allowed_roots = getattr(request.app.state, "allowed_file_roots", ())
+    if fp.is_file() and _is_allowed_file_path(fp, allowed_roots):
```
znation, line 203: file_handler calls fp.is_file() on an unresolved path — asgi_app.py:200-203
```
  fp = Path(unquote(fs_path))
  if fp.is_file() and _is_allowed_file_path(fp, allowed_roots):
```
  fp.is_file() is checked on the raw (possibly relative) path, while _is_allowed_file_path internally calls .resolve(). These two checks operate on different
  representations of the path. The check should be: resolve first, then is_file(), then the allowlist check — both operating on the same resolved Path object.

trackio/asgi_app.py
```diff
@@ -0,0 +1,231 @@
...
+            inspect.Parameter.VAR_POSITIONAL,
+            inspect.Parameter.VAR_KEYWORD,
+        ):
+            raise TrackioAPIError(f"Missing required parameter: {param.name}")
+
+    return fn(*positional_args, **keyword_args)
+
+
+async def version_handler(request: Request) -> Response:
+    package_json = Path(__file__).parent / "package.json"
```
znation: version_handler reads and parses package.json from disk on every request — asgi_app.py:124-125
```
  version = json.loads(package_json.read_text())["version"]
```
  This is a new route. The version doesn't change at runtime; cache it at module load time.

trackio/launch.py
```diff
@@ -0,0 +1,202 @@
...
+import httpx
+import uvicorn
+from uvicorn.config import Config
+
+from trackio._vendor.gradio_exceptions import ChecksumMismatchError
+from trackio._vendor.networking import normalize_share_url, setup_tunnel, url_ok
+from trackio._vendor.tunneling import BINARY_PATH
+from trackio.launch_utils import colab_check, is_hosted_notebook
+
+INITIAL_PORT_VALUE = int(os.getenv("GRADIO_SERVER_PORT", "7860"))
```
znation, line 21: These names come from the old Gradio integration and are being carried into a brand-new file that's supposed to replace Gradio. Keeping them maintains
  backward compatibility for existing users, which may be intentional — but it should be a conscious choice, and the error message on line 1738 ("Set
  GRADIO_SERVER_PORT or pass server_port to trackio.show()") would benefit from also mentioning the new expected name if these are being renamed eventually.

trackio/mcp_setup.py
```diff
@@ -0,0 +1,155 @@
...
+        trackio_server.get_settings,
+        description="Return Trackio dashboard settings and asset configuration.",
+        structured_output=True,
+    )
+
+    @mcp.tool(
+        description="Delete a run. On Spaces, pass an hf_token with write access.",
+        structured_output=True,
+    )
+    def delete_run(
```
znation, line 111: MCP mutation tools accept credentials as plain tool parameters — mcp_setup.py:110-132

  delete_run, rename_run, and trigger_sync accept hf_token and write_token as ordinary positional tool arguments. Many MCP clients log tool call invocations
  (including arguments) for debugging. Passing credentials this way risks them appearing in plain text in logs. Requiring auth to be conveyed at
  transport/connection level (e.g. HTTP headers on the /mcp mount) rather than per-call would be cleaner and more secure.

trackio/_vendor/networking.py
```diff
@@ -0,0 +1,81 @@
...
+    address = tunnel.start_tunnel()
+    return address
+
+
+def url_ok(url: str) -> bool:
+    try:
+        for _ in range(5):
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore")
+                r = httpx.head(url, timeout=3, verify=False)
```
znation, line 70: Do we really want verify=False here? (Perhaps an SSL verification error should result in a URL-not-ok?)

znation (commented):
I only skimmed as a human and generally this looks good. Some specific comments added from my Claude-powered review.

trackio/remote_client.py
abidlabs: fixed

trackio/asgi_app.py
abidlabs: agreed I don't think this will help

trackio/mcp_setup.py
abidlabs: done

trackio/asgi_app.py
abidlabs: done

trackio/launch.py
abidlabs: will keep this for backwards compatibility

trackio/mcp_setup.py
abidlabs: Yes probably, will do this in a followup since mcp usage is quite low I feel

trackio/_vendor/networking.py
abidlabs: This is the same behavior we had in gradio, I think it's fine

trackio/asgi_app.py
abidlabs: yep cached now

abidlabs:
Thanks so much Zach + Claude for the review! Mostly everything should be addressed now

znation (approved):

abidlabs:
Thanks so much for the review @znation!

PR #496, opened 2026-04-11
Prevent trackio errors from crashing the user's training loop

abidlabs:
- hardens `trackio.init()` against corrupted/malformed sqlite metadata reads and Space/bootstrap metadata failures by falling back to safe defaults instead of raising into the caller
- hardens the local and remote sender threads so sqlite/media/upload failures are converted into nonfatal warnings instead of killing the thread or escaping into user code
- hardens local fallback buffering, artifact queueing, and `finish()` so shutdown/pending-buffer checks also stay nonfatal
- makes degraded-mode warnings nonfatal even when the caller promotes warnings to exceptions
- adds regression tests for corrupt-db init, local flush failure, `finish()` pending-data inspection failure, and `warnings.simplefilter("error")`

### DIFF
--- a/trackio/__init__.py
+++ b/trackio/__init__.py
@@ -89,6 +89,13 @@ _atexit_registered = False
 _projects_notified_auto_log_hw: set[str] = set()
 
 
+def _emit_nonfatal_warning(message: str, *args, **kwargs) -> None:
+    try:
+        warnings.warn(message, *args, **kwargs)
+    except Exception:
+        print(f"* Trackio warning: {message}")
+
+
 def _cleanup_current_run():
     run = context_vars.current_run.get()
     if run is not None:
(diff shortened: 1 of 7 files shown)


trackio/__init__.py
```diff
@@ -89,6 +89,13 @@
 _projects_notified_auto_log_hw: set[str] = set()
 
 
+def _emit_nonfatal_warning(message: str, *args, **kwargs) -> None:
+    try:
+        warnings.warn(message, *args, **kwargs)
```
znation: Surprised that warnings.warn would throw!

znation (approved):

trackio/__init__.py
abidlabs: This wouldn't normally happen, only if a user has explicitly enabled it:

```py
  The main case is when the user or test harness enables warnings-as-errors, for example:

  import warnings
  warnings.simplefilter("error")
```

but I really want to harden this so we can confidently say that Trackio-related code should never cause your main experiment logging code to fail

abidlabs:
Thanks @znation for reviewing!

PR #501, opened 2026-04-15
Fix SQLite corruption on bucket-mounted Spaces

abidlabs:
- When Trackio runs on an HF Space with a bucket mount (`hf-mount`), the FUSE filesystem doesn't support file locking. SQLite depends on `fcntl`/`flock` for internal consistency, so even a single process can see corruption when locks are silently no-ops.
- Detects bucket-mount environment (`TRACKIO_BUCKET_ID` + `SYSTEM=spaces`) and switches to a **single persistent connection** with `PRAGMA locking_mode=EXCLUSIVE`. SQLite grabs the lock once and never releases it — with only one connection open, there's no contention and no reliance on filesystem locking.
- Also switches `ProcessLock` to use in-memory `threading.Lock` instead of file-based locking in the same environment.

This is a much simpler alternative to the [Parquet storage backend approach](https://github.com/gradio-app/trackio/pull/500) (~110 lines changed vs ~3500) proposed by Claude Code and reviewed by myself to solve the same root cause.

cc @Wauplin @XciD for visibility

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -159,24 +159,12 @@ def test_trackio_dir_env_var(monkeypatch):
         test_path = str(tmpdir)
 
         monkeypatch.setenv("TRACKIO_DIR", test_path)
-        monkeypatch.delenv("PERSISTANT_STORAGE_ENABLED", raising=False)
         result_dir = utils._get_trackio_dir()
         assert str(result_dir) == test_path
 
         monkeypatch.delenv("TRACKIO_DIR", raising=False)
-        monkeypatch.delenv("PERSISTANT_STORAGE_ENABLED", raising=False)
         result_dir = utils._get_trackio_dir()
         assert "huggingface/trackio" in Path(result_dir).as_posix()
-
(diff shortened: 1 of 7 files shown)


tests/unit/test_utils.py
abidlabs, line 1: Unrelated change, removes `PERSISTANT_STORAGE_ENABLED` as we no longer have persistent storage on Spaces

abidlabs:
This is relatively small so I'll go ahead and merge this in so that we can test this out in live spaces

Wauplin (commented):
Nice workaround!

abidlabs:
I investigated this a bit with @XciD and found that it probably wasn't a locking issue, it was the fact that previously (before this PR), we were trying to create a new SQlite connection every request, and some of them got dropped because the SQlite lock took a long time to resolve. Now we have a fixed single persistent connection that all requests use (each request just adds to an in-memory queue). I was not able to reproduce the stronger bucket-specific claim that HF bucket mounts silently ignore locks, and in fact, in my tests the mounted bucket behaved the same as a local /tmp file

PR #505, opened 2026-04-17
Add an `id` field to `Run` which is used internally, allowing users to have multiple runs with the same run name

abidlabs:
A large PR, because I wanted to write it in a backwards-compatible way so that older Trackio projects will continue to work with this version of Trackio

Suggested by @lewtun here: https://huggingface.slack.com/archives/C08SW1X12C8/p1776191701366079?thread_ts=1776151784.757259&cid=C08SW1X12C8

Run the `examples/crash-and-resume-same-run-name.py` example to see the new behavior.

With `resume="never"` (the default), here's the behavior:

image

With `resume="allow")`, here's the behavior:

image

This PR essentially consists of 3 pieces:

* Change Trackio internal logic and testing to use run id instead of run names to get runs
* Change the Trackio frontend to use run id so that multiple runs with the same run id can be displayed and toggled correctly
* Write defensive logic in `trackio/sqlite_storage.py` so that it works both with the older and newer schema


docs/source/python_api.md
```diff
@@ -89,11 +89,15 @@ Represents a single run in a project.
...
-- **`id`**: The run name (same as `name`)
-- **`name`**: The run name
+- **`id`**: The stable run identifier used internally by Trackio
+- **`name`**: The human-readable run name. Multiple runs can share the same name.
 - **`project`**: The project this run belongs to
 - **`config`**: The run's configuration dictionary (lazy-loaded)
 
+Trackio keeps `id` stable even when multiple runs share the same `name`. For
+older SQLite databases created before `run_id` support existed, the API exposes
+`run_name` as the effective `id` for backwards compatibility.
```
abidlabs:
```suggestion
Note: Multiple runs can share the same `name`, as Trackio will use the `id` identifier to disambiguate them internally. 
```

docs/source/track.md
```diff
@@ -309,15 +309,41 @@ trackio.finish()
...
+interruption), call [`init`] again with the same project and run name, and set
+`resume="must"`:
 
 ```python
 trackio.init(project="my_project", name="my_first_run", resume="must")
 ```
 
-This will load the existing run so you can keep logging data.
+This will load the most recently created run with that name so you can keep
+logging data using the same `run_id`.
```
abidlabs:
```suggestion
This will load the most recently created run with that name so you can keep
logging data using the same `run_id`. But if you set `resume="must"`, and no previous run exists with the same name, Trackio will raise an error. 
```

trackio/frontend/src/pages/Reports.svelte
```diff
@@ -48,9 +48,10 @@
 
     loading = true;
     try {
-      const data = await getAlerts(project, null, null, null);
-      const runSet = new Set(selectedRuns);
-      allAlerts = (data || []).filter((a) => !a.run || runSet.has(a.run));
+      const alertBatches = await Promise.all(
+        selectedRuns.map((run) => getAlerts(project, run, null, null)),
+      );
+      allAlerts = alertBatches.flat();
```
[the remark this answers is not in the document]
abidlabs: This is consistent with the pattern already used for `getLogs` directly below (line 58-60), which also fetches per-run. Keeping them uniform; a batch endpoint can be introduced later if it becomes a bottleneck.

abidlabs:
This a giant PR so instead of reviewing it manually -- just a question are we okay with making this change conceptually?

abidlabs:
Note that we'll have to slightly update this PR when https://github.com/gradio-app/trackio/pull/506 is in

abidlabs:
I'm going to merge this once CI is green in because this PR touches a lot of files and will have conflicts with subsequent PRs. I won't release for a few more days just to give us more time to test everything

PR #513, opened 2026-04-17
Reduce HF Spaces 429s: polling tuning and batched metric logs API

abidlabs:
PR reduces **HTTP 429 (rate limit)** errors on **Hugging Face Spaces** when many people use the Trackio dashboard or when many runs are selected.

**Approach**

1. **Dashboard polling (`hostPolling.js`, `App.svelte`, `Metrics.svelte`, `SystemMetrics.svelte`)**
   - On `*.hf.space`, both the app-level poll and the metrics/system poll use **2 second** intervals; on localhost they stay at **1 second** for both. Documented in `docs/source/track.md` 
   - Skip polling while the browser tab is hidden.
   - Short backoff after API responses with status **429**.
   - The “Refresh metrics realtime” control applies to metrics-oriented polling where wired.

2. **Batched metric logs API**
   - **`POST /api/get_logs_batch`**: one request returns logs for multiple runs (Metrics page uses this instead of N× `get_logs`).
   - **`get_logs_batch` in `SQLiteStorage`**: single DB connection, one loop over runs.

### DIFF
--- a/.changeset/polite-tigers-drop.md
+++ b/.changeset/polite-tigers-drop.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Reduce Spaces 429s: slower polling, get_logs_batch, read cache
--- a/.changeset/polite-tigers-drop.md
+++ b/.changeset/polite-tigers-drop.md
@@ -2,4 +2,4 @@
 "trackio": minor
 ---
 
-feat:Reduce Spaces 429s: slower polling, get_logs_batch, read cache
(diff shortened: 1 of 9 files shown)


abidlabs:
Addresses Copilot review feedback: batch log fetches wrapped in try/catch (Metrics + System metrics); read-cache uses max(DB, WAL) mtime and skips caching snapshots over 4000 rows; `_subsample_metric_rows` treats `max_points < 1` as no subsampling; `get_logs_batch` / `get_system_logs_batch` validate `runs` (max 64, dict entries) and clamp `max_points` (default 1500, cap 10000).

PR #518, opened 2026-04-18
Traces in Trackio

abidlabs:
Edit: see below: https://github.com/gradio-app/trackio/pull/518#issuecomment-4284285326


sergiopaniego (commented):
Thanks for the proposal!!
One question: would it support rendering images?
possible use case: you're training with GRPO + an env (e.g., OpenEnv), and the env returns a list of images (e.g., a browser env returning screenshots). It'd be nice to render them inline with the messages

qgallouedec:
Very cool! looking forward to integrate this!

abidlabs:
> Thanks for the proposal!!
One question: would it support rendering images?
possible use case: you're training with GRPO + an env (e.g., OpenEnv), and the env returns a list of images (e.g., a browser env returning screenshots). It'd be nice to render them inline with the messages

yep can do! We already support images in tables, so we should be able to do the same here

abidlabs:
Ok based on great feedback from everyone, have updated this PR. 

Here's a basic example: `python examples/traces/basic-trace.py`

https://github.com/user-attachments/assets/cd57f703-1c34-4696-9ee4-4488e77498a6

(I've removed many of the earliers to make the UI less opinionated, thanks @adithya-s-k for the suggestion)

A more complex example including images and tool calls: `python examples/traces/complex-trace.py` 


https://github.com/user-attachments/assets/4b71a867-6ca4-4855-bb15-4af375e50da8

cc @sergiopaniego @AmineDiro

And a potential example of how to use it with TRL: `python examples/traces/trl-trace-integration.py` (cc @qgallouedec)

Any other suggestions/improvements are welcome!

trackio/sqlite_storage.py
```diff
@@ -1882,6 +1882,123 @@ def get_logs_batch(
...
+    def get_traces(
+        project: str,
+        run: str | None = None,
+        search: str | None = None,
+        sort: str | None = None,
+        limit: int | None = None,
+        offset: int = 0,
+        run_id: str | None = None,
+    ) -> list[dict[str, Any]]:
+        logs = SQLiteStorage.get_logs(project, run, max_points=None, run_id=run_id)
```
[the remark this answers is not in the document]
abidlabs: Good point — this is a real concern for very large runs, but filtering server-side is non-trivial because trace payloads are stored inline inside metric rows (no separate trace index), so SQLite has no cheap way to skip non-trace rows without a schema change. The input normalization is addressed in the follow-up commit; I'd like to defer the scan-reduction work to a dedicated PR that introduces a lightweight trace index table so pagination/sort can be pushed down to SQL.

trackio/server.py
```diff
@@ -821,6 +821,47 @@ def get_logs_batch(
...
+    search: str | None = None,
+    sort: str | None = None,
+    limit: int | None = None,
+    offset: int | None = 0,
+) -> list[dict[str, Any]]:
+    try:
+        normalized_offset = max(0, int(offset)) if offset is not None else 0
+    except (TypeError, ValueError):
+        normalized_offset = 0
+    normalized_limit: int | None
```
znation, line 845: Double-sanitization of offset/limit

  trackio/server.py:843-856 normalizes offset and limit, then passes them to trackio/sqlite_storage.py:1968-1974 which normalizes them again with identical
  logic. One layer should own this.

  Recommendation: Remove sanitization from sqlite_storage.py and let the API layer (server.py) be the sole validator. The storage layer can trust its internal
  callers.

trackio/run.py
```diff
@@ -763,33 +764,26 @@ def _process_media(self, value: TrackioMedia, step: int | None) -> dict:
...
+                "trackio.video",
+                "trackio.audio",
+            ]:
+                file_path = value.get("file_path")
+                if file_path:
+                    absolute_path = MEDIA_DIR / file_path
+                    self._queue_upload(absolute_path, step)
+                return
+            for nested in value.values():
+                self._scan_and_queue_media_uploads(nested, step)
```
znation, line 782: Recursive _scan_and_queue_media_uploads has no depth limit

  trackio/run.py:767-786 — The refactored _scan_and_queue_media_uploads now recurses into arbitrary dicts/lists. A deeply nested trace payload (or even an
  accidental circular reference via a custom __dict__) could blow the stack. The old version was bounded to exactly 2 levels of nesting (table rows → values →
  list items).

  Recommendation: Add a max_depth parameter (e.g., 10) and stop recursing beyond it. This matches the practical ceiling for trace messages.

trackio/sqlite_storage.py
```diff
@@ -1882,6 +1882,133 @@ def get_logs_batch(
...
+                candidates = value if isinstance(value, list) else [value]
+                for index, candidate in enumerate(candidates):
+                    if (
+                        not isinstance(candidate, dict)
+                        or candidate.get("_type") != "trackio.trace"
+                    ):
+                        continue
+
+                    trace_index = index if isinstance(value, list) else None
+                    trace_id_parts = [run_id or run or "run", str(step), key]
```
znation, line 1930: Trace ID collisions across runs

  trackio/sqlite_storage.py:1934-1937 — Trace IDs are constructed as run_id_or_name:step:key[:index]. When run_id is None and run is None, the fallback is the
  string "run". If two different runs are both queried with run=None, run_id=None, they'll produce identical trace IDs, causing collisions in the frontend (the
   expand/collapse toggle uses trace.id).

  The frontend in Traces.svelte:57-61 fetches traces for multiple selectedRuns, flattening them into one array. If two runs share a step number + key, the IDs
  will collide.

  Recommendation: Include the actual run name or run ID in the trace ID unconditionally (the caller always has it from the selectedRuns list), or generate a
  unique ID (e.g., hash).

trackio/frontend/src/pages/Traces.svelte
```diff
@@ -0,0 +1,556 @@
...
+      console.error("Failed to load traces:", error);
+      traces = [];
+    } finally {
+      if (requestId === loadRequestId) {
+        loading = false;
+      }
+    }
+  }
+
+  let visibleTraces = $derived.by(() => {
```
znation, line 79: Client-side search duplicates server-side search

  Traces.svelte:79-96 — visibleTraces does full client-side filtering/sorting on the loaded traces. But getTraces in api.js also passes search/sort options to
  the server. Currently loadTraces() at line 59 calls getTraces(project, run) with no options — so the server-side search/sort/pagination is never used from
  the UI. The toolbar controls only drive the client-side $derived block.

  This means the server endpoint accepts search/sort/limit/offset parameters that the frontend never sends. The two code paths (server-side in
  sqlite_storage.py and client-side in Traces.svelte) are duplicated logic that can drift.

  Recommendation: Either remove the unused server-side filtering (YAGNI) or wire it up in the frontend and remove the client-side duplicate. Given Issue 1,
  moving filtering server-side would also be the path to fixing the performance problem.

trackio/frontend/src/pages/Traces.svelte
```diff
@@ -0,0 +1,556 @@
...
+        </select>
+      </label>
+      <div class="count">0 of {traces.length}</div>
+    </div>
+    <div class="empty-state">
+      <h2>No traces match the current filters</h2>
+      <p>Try a different search query or model filter.</p>
+    </div>
+  {:else}
+    <div class="toolbar">
```
znation, line 237: Toolbar duplication in Traces.svelte

  Traces.svelte:175-189 and Traces.svelte:198-213 — The toolbar markup (search input, sort dropdown, count display) is duplicated verbatim in both the "no
  matching traces" and "has traces" branches. If you change one, you'll need to change the other.

  Recommendation: Extract the toolbar into its own {#snippet} or move it above the conditional so it renders once regardless of whether traces match.

znation (approved):
Overall looks good (I only skimmed, Claude reviewed more thoroughly). Left some optional comments for issues that Claude found.

abidlabs:
Thanks so much for the review @znation! Cleaned up the frontend based on your comments, will merge this in once CI is green

PR #521, opened 2026-04-20
Add additional support for autonomous ML experiments

Saba9:
## Short description
Apologies for the large PR. About half the additions are from tests and docs.

  This PR adds features designed for use in AI-driven training loops where an agent or automated script needs to monitor metric health, compare runs, and query results
  programmatically without human supervision.

  **Metric watchers** (`trackio.watch()` / `trackio.should_stop()`): Register rules upfront and have Trackio fire alerts automatically on every `trackio.log()` call. Supported
  conditions:

  - `NaN`/`Inf` detection, max/min threshold breaches
  - Spike detection against a moving average
  - Stagnation — no improvement for `N` steps
  - Custom conditions via an `fn` callback: `trackio.watch("loss", fn=my_check)`

  Any condition that warrants stopping sets `trackio.should_stop()` to `True`, which training loops can poll to exit early. Custom conditions can signal early stopping by including `"stop": True` in a returned alert dict.

  **Run status tracking**: Runs now record a lifecycle status (`running` → `finished` / `failed`) in SQLite. Status is set to `running` on `init()`, `finished` on `finish()`, and `failed` on unexpected process exit. The `best`, `compare`, and `summary` commands default to finished runs only; `--include-all` overrides this.

  **New CLI commands**:

  - `trackio best --project X --metric Y [--direction min|max] [--mode last|min|max]` — find the best run for a metric
  - `trackio compare --project X [--runs ...] [--metrics ...]` — side-by-side run comparison
  - `trackio summary --project X [--runs ...]` — per-run metric summary table

  **Python API additions on `Run`**: `status`, `final_metrics`, `metrics()`, `history()`.

  **`AlertReason` constants**: Every watcher-generated alert includes a `data["reason"]` field matching one of the `trackio.AlertReason` constants, allowing agents to identify alert types programmatically. Custom conditions that return `True` use `AlertReason.CUSTOM`.

  ## AI Disclosure

  - [x] I used AI to write everything. I then manually reviewed and revised the code.

  ## Type of Change

  - [x] Documentation update

  ## Related Issues

  Closes:

  ## Testing and linting

  ```bash
  python -m pytest


znation:
I had Claude analyze it and write up some actionable feedback. At first I was going to go through and leave comments for each item in the source code, but I realized it's probably more efficient to just include the Markdown file here. 
[autonomous-ml-trackio-pr-review.md](https://github.com/user-attachments/files/27395623/autonomous-ml-trackio-pr-review.md)

Note that Claude considers two issues blockers:

```
B1. Run-status writes hit the local SQLite DB unconditionally — even for Space / self-hosted runs
B2. Remote-mode `Run.alert(..., data=...)` silently drops the `data` field
```

Saba9:
I'm going to close this for now. I think this PR would be better split into multiple more tightly scoped PRs.

PR #525, opened 2026-04-20
Restore sidebar share and embed snippets, and fix query parameter regression

abidlabs:
- restore the old sidebar "Share this view" and "Embed this view" UI in the Svelte sidebar when Trackio is deployed to Spaces
- generate share URL and iframe embed snippet from current state (`project`, `metrics`, selected runs, and `accordion` visibility) and fix a bug where these query parameters were not being applied. Add a test to prevent this regression from happening

Try it out here: https://huggingface.co/spaces/abidlabs/trackio-on-spaces-702221


abidlabs:
Will merge this in so that we can get a patch release out shortly

PR #531, opened 2026-04-22
Add configurable custom frontends for Trackio

### LINKED ISSUE #491: Allow easily vibecoding custom frontends for Trackio logs

abidlabs:
Closes #491

- add configurable custom frontend resolution for Trackio with CLI arg, Python API arg, env var, and persistent user config
- serve arbitrary static frontend directories locally and deploy the selected frontend to Gradio and static Spaces

Usage:

1. Run `trackio show --frontend ./my-trackio-frontend`.
2. Ask your LLM to edit the files in that directory.
3. Keep the browser open while Trackio live reloads the frontend as those files change.

If the directory passed to `--frontend` does not exist, or exists but is empty, Trackio copies in the starter frontend automatically, prints that it did so, and then serves that directory. The starter is a complete plain-HTML/CSS/JS template: it calls the Trackio API, loads projects and runs, fetches metric values, and draws simple charts that you can replace with your own UI.

Example Frontends:

### Signal Console
Signal Console

### Sunrise Cards
Sunrise Cards

### Editorial Grid
Editorial Grid


qgallouedec:
nice! can we also set e.g. the width of the plots?

abidlabs:
> nice! can we also set e.g. the width of the plots?

For sure, the frontend is just vanilla js/css/html, so you can modify anything now. I'll vary up these 4 examples a bit more so we can see what's possible

abidlabs:
PR should be ready now!

abidlabs:
Going to do a release to get the Windows fix out. Will go ahead and merge this in (I've tested this fairly robustly & it doesn't interact much with core Trackio library so its unlikely to have cause issues)

PR #540, opened 2026-04-29
Raise metrics sampling cap from 1500 to 3000

edbeeching:
## Summary
Bumps the default \`max_points\` cap used by \`get_logs\` / \`get_logs_batch\` (and the system-logs equivalents) from 1500 to 3000. Client-side smoothing on the Metrics tab now runs over twice as many samples on longer runs, reducing index-stride subsampling artifacts. 3000 is comfortably below the existing \`_LOGS_BATCH_MAX_POINTS = 10_000\` hard ceiling, so wire size and chart-render cost stay bounded.

The original trace-viewer + step-slider work from this PR has been superseded by #559 and #560, so this branch is now narrowed to the one piece that wasn't shipped: the metrics sampling resolution bump.

## Test plan
- [x] \`pytest tests/unit\`
- [x] \`ruff check --fix --select I && ruff format\`

### DIFF
--- a/.changeset/raise-metrics-cap.md
+++ b/.changeset/raise-metrics-cap.md
@@ -2,4 +2,4 @@
 "trackio": patch
 ---
 
-perf: raise default metrics sampling cap from 1500 to 3000 so client-side smoothing on the Metrics tab runs over higher-resolution data
+fix: raise default metrics sampling cap from 1500 to 3000 so client-side smoothing on the Metrics tab runs over higher-resolution data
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @edbeeching for opening this PR! A simpler approach would be to just store traces in a separate sqlite table than the scalar metrics. I opened a separate PR to do that: https://github.com/gradio-app/trackio/pull/559

I'll keep this PR open for now since some of the other fixes in this PR are still relevant and we might cherrypick and merge those in.

abidlabs (approved):

abidlabs:
Thanks @edbeeching moved most of the changes in this PR to #559 and #560, but will keep the fix for the smoothing issue you noticed here, and merge this in when CI passes.

PR #548, opened 2026-05-18
fix: preserve sparse metrics during row subsampling to prevent incorrect bar plots (#543)

### LINKED ISSUE #543: [BUG] Epoch chart is rendered as a filled rectangle instead of a line
Hi, I noticed an issue with how the `epoch` metric is displayed in the chart UI.

# Expected behavior

The `epoch` chart should be rendered as a normal line chart, similar to other scalar metrics. Since `epoch` is logged as a scalar value over time, I expected it to appear as a line or points.

# Actual behavior

The `epoch` chart is rendered as a large filled rectangle instead of a line and trackio only record last epoch.
However, the per-step `train/loss` chart is displayed correctly.

# Screenshot

Image

# Code

```python
for epoch in epochs:
    total_loss = 0.0

    for X, y in dataloader:
        loss = loss_fn(model(X), y)
        total_loss += loss.item()
        trackio.log({"train/loss": loss.item()}, step=epoch)

    avg_loss = total_loss / len(dataloader)
    metrics = {
        "epoch": epoch,
        "train/avg_loss": avg_loss,
    }
    trackio.log(metrics, step=epoch)
```

mahadrehmann:
Thank you for your contribution! All PRs should include the following sections. PRs missing these sections may be closed immediately.

## Short description

This PR fixes a bug where sparse time-series metrics (like epoch) were being incorrectly rendered as a filled rectangle or a single-point bar chart instead of a continuous line chart.

### Root Cause:
Previously, _subsample_metric_rows() utilized a blind uniform interval sweep to keep the payload under the max_points limit. Because sparse metrics are logged much less frequently than dense metrics (like train/loss), they were being completely dropped by this sweep.

### Solution:
The new implementation adds a safe “Base‑Sample + Injection” strategy that preserves sparse metrics while keeping the total payload strictly under the max_points limit:

1. Perform the base uniform sample using max_points to preserve the overall timeline footprint.
2. Group the rows by the metrics they contain.
3. If a metric is severely underrepresented in the base sample (defined as < 10 points), explicitly inject a small, uniform subsample of its indices (capped at 50) to guarantee the frontend receives enough data to render a continuous line chart.

## AI Disclosure

We encourage the use of AI tooling in creating PRs, but the any non-trivial use of AI needs be disclosed. E.g. if you used Claude to write a first draft, you should mention that. Trivial tab-completion doesn't need to be disclosed. **You should self-review all PRs, especially if they were generated with AI**. 

-----

- [x] I used AI to... write the initial draft of the _subsample_metric_rows refactor and help identify edge cases (like preventing payload explosions if hundreds of metrics are logged). I have manually reviewed and tested the final logic.
- [ ] I did not use AI

----

## Type of Change

- [x] Bug fix
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

If this PR closes an issue, please link it below: 

Closes: #543 

## Testing and linting

   ```bash
   python -m pytest
   ```

   ```bash
   ```

### DIFF
--- a/tests/unit/test_sqlite_storage.py
+++ b/tests/unit/test_sqlite_storage.py
@@ -500,3 +500,63 @@ def test_query_project_normalizes_bytes(temp_dir):
 def test_query_project_missing_project(temp_dir):
     with pytest.raises(FileNotFoundError):
         SQLiteStorage.query_project("nonexistent", "SELECT 1")
+
+
+def _make_row(metrics_dict):
+    """Helper: create a fake row dict as returned by SQLiteStorage queries."""
+    return {"metrics": orjson.dumps(metrics_dict)}
+
+
+def test_subsample_preserves_sparse_metrics():
+    """Sparse metrics logged every N rows must survive aggressive subsampling.
+
+    Reproduces the bug reported in #543: when a dense metric (train/loss) is
+    logged every row and a sparse metric (epoch) is logged every 500 rows, the
+    old blind-sweep subsampler dropped nearly all epoch rows, causing the
+    frontend to treat epoch as a single-point bar chart instead of a line chart.
+    """
+    max_points = 150
+    # epoch appears every 500 rows – 10 times in 5 000 rows
+    rows = []
+    for step in range(5000):
+        m = {"train/loss": 1.0 / (step + 1)}
+        if step % 500 == 0:
+            m["epoch"] = step // 500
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @mahadrehmann, I fixed this in https://github.com/gradio-app/trackio/pull/550

PR #559, opened 2026-05-19
Store traces separately from metrics

abidlabs:
## Summary
- Add a dedicated SQLite `traces` table and split `Trace` payloads out of metric rows at write time.
- Make `get_traces` query the traces table directly with backend search, sorting, limit, and offset instead of loading all metric logs and slicing in Python.
- Export/import traces through parquet and teach the static frontend API to read `aux/traces.parquet`.
- Keep scalar metric logs free of large trace payloads by construction.

## Validation
- pytest tests/unit/test_trace.py -q
- pytest tests/unit/test_sqlite_storage.py -q
- python -m compileall -q trackio
- npm --prefix trackio/frontend run lint
- npm --prefix trackio/frontend run build

### DIFF
--- a/.changeset/thick-vans-vanish.md
+++ b/.changeset/thick-vans-vanish.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat: Store traces separately from metrics
--- a/.changeset/thick-vans-vanish.md
+++ b/.changeset/thick-vans-vanish.md
@@ -2,4 +2,4 @@
 "trackio": minor
 ---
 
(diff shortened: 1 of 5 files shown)


abidlabs:
Going to merge this in so that we can take care of the remaining todos from https://github.com/gradio-app/trackio/pull/540:

- Reworks the Traces tab into a dataset-viewer-style table with page-sized loading, run selection, URL state, column toggles, search, sorting, and inline expandable cells.
- Adds a step slider/count view so large batched trace runs can load one training step at a time.

PR #563, opened 2026-05-21
Fix Traces UI a bit

abidlabs:
- show relevant part of trace ID
- shortern the trace ID column

Before:

Screenshot 2026-05-21 at 10 48 32 AM

After:

image


abidlabs:
Or should I just remove the Trace ID from the table view altogether? Not sure its useful actually

PR #569, opened 2026-06-01
add viewer (lightbox) to Media & Tables section

catwell:
## Short description

This PR adds a filter and an image viewer (lightbox).

I "wrote" (see AI Disclosure) this because I need it. I work on image generation models so I spend a lot of time on that tab. This lets me select a specific sample or set of samples with the filter and browse through steps with my keyboard arrows.

Screenshot From 2026-06-01 21-29-31
Screenshot From 2026-06-01 21-30-10

## AI Disclosure

- [x] I used AI to write this whole thing.
- [ ] I did not use AI

This is basically a single prompt in Claude Code / Opus 4.8. I did read what came out and it looks reasonable to me but I am not a front-end developer and I do not use Svelte.

----

## Type of Change

- [ ] Bug fix
- [ ] Documentation update
- [ ] Test improvements

### DIFF
--- a/trackio/frontend/src/pages/Media.svelte
+++ b/trackio/frontend/src/pages/Media.svelte
@@ -3,6 +3,7 @@
   import WaveformAudio from "../components/WaveformAudio.svelte";
   import { getLogs, getMediaUrl, isStaticMode, fetchMediaBlob } from "../lib/api.js";
   import { buildColorMap } from "../lib/stores.js";
+  import { filterMetricsByRegex } from "../lib/dataProcessing.js";
 
   let {
     project = null,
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @catwell let me see if we can adapt this PR a bit and include it as part of Trackio

catwell:
Something I forgot to add: I use filterMetricsByRegex to filter something that is not a metric. But that function is actually generic and could be renamed to just filterByRegex maybe. I did not do it to keep the diff contained to a single file.

abidlabs:
Thanks @catwell! We liked this and included a version of this here: https://github.com/gradio-app/trackio/pull/579

PR #571, opened 2026-06-03
Fix IndexError when importing CSV with no numeric metric columns

Ckal:
## Problem
`import_csv()` guards `SQLiteStorage.bulk_log()` with `if metrics_list:`, but then calls `metrics_list[0].keys()` unconditionally two lines later. When a CSV has a valid `step` column but no parseable numeric metric columns, `metrics_list` is empty — the import prints a misleading `* Imported 0 rows` and then crashes with `IndexError: list index out of range`.

## Repro
```python
# logs.csv:
# step,timestamp,note
# 1,2024-01-01T00:00:00+00:00,start
import trackio
trackio.import_csv("logs.csv", project="demo")  # IndexError

### DIFF
--- a/.changeset/fresh-rocks-lead.md
+++ b/.changeset/fresh-rocks-lead.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Fix IndexError when importing CSV with no numeric metric columns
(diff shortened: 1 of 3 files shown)


abidlabs (approved):
LGTM thanks @Ckal

PR #581, opened 2026-06-10
fix: limit default run selection to most recent N runs on open

### LINKED ISSUE #573: Only select at most N runs on open
When you have many runs in a project, they all get selected and Trackio gets very slow.

To avoid this I limit the number of runs selected this way:

Image

I can send a PR if you want (tests need to be modified as well).

KomalDeep355:
Fixes #573

## Problem
When a project has many runs, all of them are selected on open, which makes the Trackio dashboard very slow.

## Solution
Added a 'MAX_SELECTED_RUNS = 10' constant and a 'max_runs' parameter to 'get_runs_for_project()'. Only the most recent N runs are selected by default. All runs remain visible — only the default Selection is capped.

## Changes
- Added 'MAX_SELECTED_RUNS = 10' constant in 'server.py'
- Updated 'get_runs_for_project()' to accept and apply 'max_runs'

### DIFF
--- a/trackio/server.py
+++ b/trackio/server.py
@@ -680,8 +680,16 @@ def get_metric_values(
     )
 
 
-def get_runs_for_project(project: str) -> list[dict[str, Any]]:
-    return SQLiteStorage.get_run_records(project)
+MAX_SELECTED_RUNS = 10
+
+
+def get_runs_for_project(
+    project: str, max_runs: int | None = MAX_SELECTED_RUNS
+) -> list[dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @KomalDeep355 but I think a better fix would be to solve the undelrhying issue of Trackio becoming slow, will look into it

KomalDeep355:
Thanks for the feedback @abidlabs! That makes sense — fixing the root cause is definitely the better approach. I'd love to help with that if possible. Could you point me to where the slowness originates in the codebase? Happy to take a look and submit a follow-up PR.

PR #582, opened 2026-06-10
Add JSONL inbox fragments: durable logging when the Space is unreachable and on network filesystems

### LINKED ISSUE #544: Trackio logs may never sync if Space is unavailable at run end (no retry unless next init)
## Description

When using Trackio with Hugging Face Trainer, logs are not reliably synced to the remote Space if the Space is unavailable during or at the end of training. In particular, if the process exits and no subsequent trackio.init() is called, locally buffered logs may never be uploaded.

This is problematic for workflows where:

- training runs are ephemeral (e.g., spot instances)
- The machine is not reused after a run
- users expect logs to be persisted remotely automatically

## Reproduction

```
* Trackio project initialized: huggingface
* Trackio metrics will be synced to Hugging Face Bucket: https://huggingface.co/buckets/madoss/trackio-bucket
* Found existing space: https://huggingface.co/spaces/madoss/trackio
/home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/trackio/utils.py:27: UserWarning: trackio.init() could not create a remote client for Space 'madoss/trackio': Could not connect to Space 'madoss/trackio'. Is it running?
Client.__init__() got an unexpected keyword argument 'hf_token'. Continuing with local fallback metadata lookups.
  warnings.warn(message, *args, **kwargs)
/home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/trackio/utils.py:27: UserWarning: trackio.init() could not inspect existing runs for project 'huggingface' on Space 'madoss/trackio': Could not connect to Space 'madoss/trackio'. Is it running?
Client.__init__() got an unexpected keyword argument 'hf_token'. Continuing without resume metadata.
  warnings.warn(message, *args, **kwargs)
* Created new run: test-infra-dry-run
{'eval_loss': '3.47', 'eval_bleu': '2.386', 'eval_chrf': '15.38', 'eval_runtime': '7.144', 'eval_samples_per_second': '1.4', 'eval_steps_per_second': '0.42', 'epoch': '0.3125'}                                                                                                                                    
{'eval_loss': '3.338', 'eval_bleu': '3.782', 'eval_chrf': '19.21', 'eval_runtime': '6.668', 'eval_samples_per_second': '1.5', 'eval_steps_per_second': '0.45', 'epoch': '0.625'}                                                                                                                                    
{'eval_loss': '3.312', 'eval_bleu': '5.843', 'eval_chrf': '18.91', 'eval_runtime': '10.05', 'eval_samples_per_second': '0.995', 'eval_steps_per_second': '0.299', 'epoch': '0.9375'}                                                                                                                                
{'eval_loss': '3.253', 'eval_bleu': '3.652', 'eval_chrf': '19.1', 'eval_runtime': '7.262', 'eval_samples_per_second': '1.377', 'eval_steps_per_second': '0.413', 'epoch': '1.25'}                                                                                                                                   
{'eval_loss': '3.512', 'eval_bleu': '3.464', 'eval_chrf': '18.81', 'eval_runtime': '9.786', 'eval_samples_per_second': '1.022', 'eval_steps_per_second': '0.307', 'epoch': '1.562'}                                                                                                                                 
{'train_runtime': '57.93', 'train_samples_per_second': '6.905', 'train_steps_per_second': '3.452', 'train_loss': '3.285', 'epoch': '1.562'}               
 25%|█████████████████████████████                                                                                       | 50/200 [00:55<00:59,  2.51it/s]* Run finished. Uploading logs to the remote Trackio server (please wait...)
/home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/trackio/utils.py:27: UserWarning: * Some logs could not be sent to the remote server (it may still be starting up). They have been saved locally and will be sent automatically next time you call: trackio.init(project="huggingface", space_id="madoss/trackio")
```

## Additional Notes

There also seems to be a related warning:

```
Client.__init__() got an unexpected keyword argument 'hf_token'
UserWarning: trackio.init() could not create a remote client for Space 'madoss/trackio': Could not connect to Space 'madoss/trackio'. Is it running?
```

This may be related to #534 ?

## Environment

- Python: 3.12
- `trackio==0.25.1, transfomers==5.5.4, huggingface-hub==1.11.0`

### LINKED ISSUE #554: Architecture change: make the Bucket the source of truth, Spaces should always be static
The idea is for the Trackio Client to write **directly logs to Buckets** (in Parquet format) instead of going through the Space and Spaces to always be static and only read from this Parquet Bucket. Besides the significant code deduplication, this would have several benefits as discussed below:

## The Real Motivation

Today writes go through the Space's HTTP API (`/bulk_log`), and the Space owns a SQLite file that flushes to Bucket periodically. The Space being up is on the critical path for log durability. That causes real problems:

- If the Space is paused/cold/broken at the end of a training run, logs sit in a local pending buffer and only drain on the next `trackio.init()` for that project (#544). On ephemeral infra (spot, one-shot CI, scratch boxes) the next `init()` never happens and the data is lost with the machine.
- Cold starts (~1 min on free hardware) and the 30s `finish()` join often miss each other, so we ship users a warning that boils down to "your logs might never sync."

Buckets are the surface HF built for this workload. They don't carry the `<100k files`, `<10k per folder`, `<100 per commit` recommendations that bite git-based repos, and there's no documented per-second cap. Going bucket-first is strictly better on rate limits, and removes the Space as a write dependency.

## Proposed shape

**Bucket = canonical store. Space = read-cache + UI + alert/webhook evaluator.**

Clients write append-only parquet shards directly to the bucket:

```
<project>/runs/<run_id>/
  meta.json
  shards/<writer_id>/<seq>.parquet
  system/<writer_id>/<seq>.parquet
  media/<sha256>.<ext>           (content-addressed)
```

- Each client process picks a random `<writer_id>` at `init()`, owns its own subtree → no key collisions, no coordination, distributed training works for free.
- Shards are immutable, no manifest needed — list-by-prefix is enough.
- Client buffers `log()` calls in memory and flushes a shard on a time/size threshold (e.g. 10s). A training job at 10 batches/sec produces ~2.8k well-sized shards/day instead of ~864k tiny PUTs.
- Space polls shards above a per-writer watermark, merges into its local view, fires alerts/webhooks on observed new rows (dedupe by `log_id`, which is already a uuid).

## Refinement: drop SQLite entirely (local and remote)

The cleanest version of this is to not keep SQLite as a local hot view at all — use the same parquet-shard layout for both local and Bucket-backed projects, queried via DuckDB. Local mode points the writer at `~/.cache/huggingface/trackio/<project>/`, remote at a Bucket; identical code path. That collapses local and remote onto one storage primitive, deletes the connection/journal/locking plumbing in `sqlite_storage.py`, and makes cluster filesystems (Lustre, FSx, GPFS, NFS, WekaFS) safe by construction rather than by carefully-chosen PRAGMAs — see #555. Costs: DuckDB as a runtime dep (~10 MB, replaces stdlib `sqlite3`), mutations reshape slightly (`delete_run` = remove a directory, `rename_run` = rewrite `meta.json`), and a one-shot migration from existing `~/.cache/huggingface/trackio/*.db` files.

### LINKED ISSUE #555: Ensure Trackio works on weird cluster filesystems (Lustre, FSx, GPFS, NFS, WekaFS)
## Context

[Reported via Slack](https://github.com/gradio-app/trackio/issues/555#issue-4473559792) by a user training on a cluster with `/fsx` (WekaFS) as the local Trackio cache:

> I've been hitting a consistent `SIGBUS` crash during distributed training with trackio on our cluster. I traced the python trace down to a known SQLite limitation on distributed file systems: a race condition where a WAL checkpoint calls `ftruncate` while another thread is actively reading via `mmap`. Because cache invalidation isn't atomic across the network for WekaFS, the kernel throws a SIGBUS when the mmap reader hits the truncated page.

This isn't WekaFS-specific — it's a known SQLite caveat on every distributed/network filesystem (Lustre, FSx, GPFS, NFS, CephFS, …). Combine **WAL journal mode** + **memory-mapped reads** + **non-coherent caches** and you get SIGBUS or worse. SQLite's own docs warn against running over network filesystems for exactly this reason.

The user worked around it by `--trackio_space_id ...` (logging directly to the Space, bypassing the local SQLite cache). But the local cache path is the default, and cluster filesystems are the default storage for cluster training jobs, so we'd like Trackio to be safe out of the box there.

Currently, `trackio/sqlite_storage.py:51:

- `journal_mode` defaults to `WAL` locally, `DELETE` on Spaces. Overridable via `TRACKIO_SQLITE_JOURNAL_MODE`.
- `synchronous = NORMAL`, `temp_store = MEMORY`, `cache_size = -20000`.
- `locking_mode = EXCLUSIVE` on Spaces.

What we don't set:

- `PRAGMA mmap_size` — left at SQLite's compile-time default (typically 0 on most builds, but non-zero on some). On a distributed FS, any non-zero value is unsafe.
- No detection of "this path is on a network filesystem" to auto-pick safe pragmas.
- No documentation telling cluster users which knobs to flip.

What we could do:

1. **Default `PRAGMA mmap_size=0`.** Memory-mapped I/O wins ~nothing for Trackio's workload (small, write-heavy, low-locality) and is the direct trigger for the SIGBUS class of bugs on distributed FSes. Make `0` the default everywhere; expose `TRACKIO_SQLITE_MMAP_SIZE` for users who want to opt back in.
2. **Add `TRACKIO_SQLITE_*` env knobs for the other pragmas** we care about — at least `JOURNAL_MODE` (exists), `MMAP_SIZE` (new), `SYNCHRONOUS` (new), `LOCKING_MODE` (new), `TEMP_STORE` (new). Document them in `environment_variables.md`.
3. **Document a recommended cluster setup** in the docs: either point `TRACKIO_DIR` to a node-local disk (e.g. `/tmp` or `\$SLURM_TMPDIR`), or set `TRACKIO_SQLITE_JOURNAL_MODE=DELETE` + `TRACKIO_SQLITE_MMAP_SIZE=0` if local disk isn't available. Add a one-liner heuristic if we want to be proactive (e.g. warn if `TRACKIO_DIR` resolves under a path matching `^/(fsx|lustre|gpfs|nfs|cephfs|weka)` and the user hasn't overridden the pragmas).

## Intersection with #554 (Bucket-as-source-of-truth)

The medium-term architecture in #554 makes this whole class of bugs largely irrelevant for the *write* path: clients buffer in memory and flush parquet shards to a Bucket. There is no shared SQLite file across nodes, no WAL checkpoint contending with another process's mmap reader.

## Long-term: retire the hazard class

The safe-defaults work above is still the right tactical step — a few-line change that unblocks cluster users in days. The strategic follow-up lives in #554: replace SQLite entirely with append-only parquet shards queried via DuckDB, same layout for local and Bucket-backed projects. With no WAL, no `mmap`, no `ftruncate` racing readers, the SIGBUS class of bugs on distributed filesystems goes away by construction rather than by carefully-chosen PRAGMAs. The two don't conflict — #555 buys time, #554 finishes the job.

abidlabs:
Closes: #544
Closes: #554
Closes: #555 

## What this does

Adds an append-only JSONL fragment format (`trackio/fragments.py`) used as a durable fallback in two situations, as described below. Also: safe SQLite defaults and a client-compat bugfix.

**1. The Space is unreachable (cold start, paused, broken) — #544**

The client already creates the Bucket at `init()` and holds a write token, so it can make data durable without the Space:

- **While waiting for the Space to come up** (no client yet), the client spills its buffered logs as JSONL fragments to the Bucket inbox (`trackio/inbox/<writer_id>/<seq>.jsonl`) every 30s — *without* clearing the local pending buffer, so the normal `/bulk_log` replay still happens if the Space becomes reachable (the Space dedupes by `log_id`).
- **Transient mid-run send failures keep today's exact behavior**: persist locally, replay via `/bulk_log` after recovery. The Bucket is not involved (this preserves the `test_data_robustness.py` e2e contract).
- **`finish()`** (also reached via the existing `atexit` handler) force-drains anything still pending to the Bucket and clears it, instead of only printing "will be sent next time you call trackio.init()". On ephemeral infra (HF Jobs, spot instances), data now survives even if the Space never became ready during the job. When the Space keeps failing at shutdown and a Bucket is available, the sender thread exits promptly instead of retrying through the full 30s join.
- Pending media uploads are uploaded straight to their final Bucket location (`trackio/media/...`), which existing Spaces already serve from the `/data` mount.
- The Space imports inbox fragments at startup (`SQLiteStorage.load_from_dataset`) and via a background poller (default every 15s, `TRACKIO_INBOX_POLL_INTERVAL`); the dashboard also imports the local inbox synchronously at server startup, then deletes consumed fragments from the Bucket. `force_sync` also triggers an import.

**2. `TRACKIO_DIR` is on a network filesystem (NFS, Lustre, FSx, GPFS, WekaFS, CephFS, ...) — #555**

New `TRACKIO_STORAGE_MODE` env var (`auto`|`sqlite`|`jsonl`, default `auto`):

- In `jsonl` mode, training processes never write SQLite. Local runs write fragments to `TRACKIO_DIR/inbox/`, one file per process (atomic tmp+rename); the `trackio show` server imports them, making the dashboard the **single SQLite writer**. This removes the WAL-checkpoint/mmap/multi-writer hazard class (SIGBUS etc.) by construction.
- `auto` detects network filesystems from `/proc/mounts` (Linux only) and switches to `jsonl` with a printed notice.

### Scenarios

| Scenario | Outcome |
|---|---|
| Space cold at job end, machine destroyed | Fragments spilled to Bucket every 30s while waiting + final drain in `finish()`/atexit; Space imports them on boot |
| Job killed mid-run during Space cold start | Everything up to the last 30s spill window is already in the Bucket |
| Transient network error, then recovery | Exactly today's behavior: persist locally, replay via `/bulk_log`; Bucket untouched |
| No network at all | Local pending buffer, same as today; drains on next send/init |
| Space healthy throughout | No fragments written; hot path unchanged |
| Fragment imported twice / also replayed via `/bulk_log` | Deduped by `log_id` |
| Space crashes between import and fragment delete | Re-import is idempotent |
| SQLite on Lustre/WekaFS/NFS | `jsonl` mode: training processes never open the DB for writes; `mmap_size=0` everywhere regardless |

###  Changes in behavior / usage

**No breaking changes for the public API** (`trackio.init/log/finish` signatures unchanged). Things reviewers and users should know:

1. **New Bucket layout prefix**: clients may write `trackio/inbox/**.jsonl` to the project Bucket; the Space deletes these after import. Anything consuming the Bucket layout directly should ignore this prefix. (On Spaces running an older Trackio build, fragments sit unimported in the inbox until the Space rebuilds with a newer version — data is delayed, not lost; live-client replay via `/bulk_log` is unaffected.)
2. **`Run.__init__` gained a `bucket_id` keyword arg** (inserted after `space_id`). Anyone constructing `Run` directly with positional args past `space_id` must update; `trackio.init()` callers are unaffected.
3. **`finish()` messaging changed**: when the Space is unreachable but the Bucket write succeeds, the previous warning ("saved locally, will be sent next time you call trackio.init()") is replaced by a message that logs were uploaded to the Bucket.
4. **Alerts for remote runs** are now written to the Bucket inbox when the Space send fails (previously they were stored in local SQLite with no replay mechanism, i.e. effectively lost for the Space).
5. **New env vars**: `TRACKIO_STORAGE_MODE`, `TRACKIO_INBOX_POLL_INTERVAL`, and the `TRACKIO_SQLITE_*` pragma overrides (documented in `docs/source/environment_variables.md`).
6. **SQLite default change**: `mmap_size` is now explicitly `0` on all connections (previously left at the build default). Opt back in via `TRACKIO_SQLITE_MMAP_SIZE`.
7. **`jsonl` mode tradeoffs** (only when opted-in or auto-detected on a network FS):
   - Locally logged metrics become visible only after the dashboard server imports the inbox (at startup, every poll interval, and on `force_sync`). With no dashboard running, data sits as fragments in `TRACKIO_DIR/inbox/` until one starts.
   - Resuming a run reads the last step from SQLite, so steps may restart from 0 if fragments haven't been imported yet.
   - Media for failed remote uploads goes directly to the Bucket; there is no SQLite staging of uploads.
8. **Dashboard/Space now runs a background inbox poller thread** (daemon, started in `build_starlette_app_only`).

9. **`init()` no longer blocks up to 300s waiting for a newly created Space to build** — it now waits at most 15s, then proceeds with a note that the Space is still building. The Bucket inbox makes blocking unnecessary: logging starts immediately, a live client delivers via `/bulk_log` replay the moment the Space is up, and fragments cover the case where the process exits first. Fast build failures (`NO_APP_FILE`, `CONFIG_ERROR`) are still surfaced within the window. `trackio.sync()` keeps the full 300s wait since it must push to a RUNNING Space.

### How I verified it

I ran two identical HF Jobs that each logged 5 metrics to a Space and exited immediately (machine destroyed). The only difference is the trackio version installed in the job:

- 🔴 **trackio @ `main`** → https://huggingface.co/spaces/abidlabs/trackio-demo-main-10b8e8 — dashboard is empty. The job ended with *"saved locally and will be sent automatically next time you call trackio.init()"* on a machine that no longer exists, so the data is gone.
- 🟢 **trackio @ this PR** → https://huggingface.co/spaces/abidlabs/trackio-demo-pr582-10b8e8 — the `preempted-job-demo` run is fully there (5 logged steps + config). `finish()` uploaded a JSONL fragment to the Bucket inbox, and the Space imported it when it woke up.


abidlabs:
Will do some more testing and try to merge this in tomorrow so that we can let it stew for a bit on `main` before the release

abidlabs:
I tested and confirmed that both scalar metrics and media are logged correctly and no data is lost even if the Space is not ready yet. (Test script: `examples/deploy-images-on-spaces.py`)

Produces: https://huggingface.co/spaces/abidlabs/deploy-images-on-spaces-157998

abidlabs:
I'll go ahead and merge this in when CI finishes running

PR #586, opened 2026-06-18
Add artifact tracking: versioned file bundles with log_artifact / use_artifact

### LINKED ISSUE #546: Artifact tracking
Hello,
Thank you for Trackio, it’s a great tool.

Is artifact tracking on the roadmap? It would be an amazing addition and would cover the rest of our experiment tracking needs.

The methods that could be covered first:
- `wandb.Artifact` + `add_file` / `add_dir` / `add_reference`
- `run.log_artifact` / `run.use_artifact`

Thank you!

Saba9:
## Short description

This PR adds **artifact tracking** to Trackio — a versioned, named bundle of files (a model checkpoint, a dataset, eval outputs) attached to a project, with a wandb-compatible Python API.

**API surface**
- `trackio.Artifact(name, type, description, metadata)` with `add_file()` / `add_dir()` and `download()`
- `trackio.log_artifact(artifact_or_path, ...)` — log files as an **output** of the current run (accepts an `Artifact` or a bare file/dir path)
- `trackio.use_artifact("name[:version|:alias]", ...)` — fetch an artifact and record it as an **input**, building run↔artifact lineage

**What it does**
- **Versioning & dedup** — content-addressed storage (new `trackio/cas.py`); each log under a name records the next version (`v0`, `v1`, …), and identical content reuses the existing version rather than creating a new one.
- **Aliases** — moving pointers (`latest` auto-assigned, plus user aliases like `prod`/`best`) that resolve to a single version; version specifiers are reserved.
- **Lineage** — runs are linked to the artifacts they produce and consume.
- **Local-first + remote** — works fully offline against SQLite; when a run targets a Hugging Face Space or self-hosted server, blobs are uploaded once (skipping blobs the server already has) and synced to the configured HF Dataset / bucket. `download()` is idempotent and fetches missing blobs from the remote with authentication.

**Implementation notes**
- New SQLite tables: `artifacts`, `artifact_versions`, `artifact_aliases`, `run_artifact_links`, committed atomically.
- New server routes for blob presence checks, bulk blob upload, artifact logging, manifest fetch, and use; untrusted manifest fields are validated and logical paths are guarded against traversal.
- New modules: `trackio/cas.py`, `trackio/pending_uploads.py`, `trackio/artifact.py`.
- Docs: new `docs/source/artifacts.md` guide.

**Note:** This PR covers the API, storage, sync, and docs. **The dashboard UI for browsing artifacts will be added in a follow-up PR.**

## AI Disclosure

- [x] I used AI to implement the artifact tracking feature. I self-reviewed the changes.
- [ ] I did not use AI

## Type of Change

## Related Issues

Closes: #546

## Testing and linting

New coverage under `tests/unit/`: `test_artifact.py`, `test_artifact_server.py`, `test_artifact_remote.py`, `test_artifact_remote_e2e.py`, `test_artifact_download_remote.py`, `test_artifact_bucket_sync.py`, `test_artifact_dataset_sync.py`, `test_artifact_e2e.py`.


abidlabs:
Awesome feature @Saba9, will work great with Buckets. Will review this PR

trackio/run.py
```diff
@@ -1153,6 +1191,229 @@ def log(self, metrics: dict, step: int | None = None):
...
+            else:
+                artifact.add_file(path)
+
+        if artifact._logged:
+            raise RuntimeError(
+                "Artifact has already been logged or fetched; "
+                "construct a new Artifact() to log again."
+            )
+
+        user_aliases = list(aliases) if aliases else []
```
abidlabs: `aliases` needs type/content validation before converting with `list()`. As written, `aliases="prod"` becomes four aliases (`p`, `r`, `o`, `d`), and `aliases=[""]` is accepted and persisted even though `use_artifact("m:")` explicitly rejects an empty version/alias so the alias cannot be addressed through the public API. I would require `aliases` to be a list/tuple of non-empty strings, validate each alias with the same name grammar (and reserve `vN`), and mirror that validation in the server endpoint for remote/direct clients.

trackio/sqlite_storage.py
```diff
@@ -3828,41 +4082,565 @@ def add_pending_upload(
...
+        version_id: int,
+        direction: str,
+        now: str,
+    ) -> None:
+        if direction not in ("input", "output"):
+            raise ValueError(
+                f"direction must be 'input' or 'output', got {direction!r}"
+            )
+        conn.execute(
+            """INSERT OR IGNORE INTO run_artifact_links
```
abidlabs, line 4438: This `INSERT OR IGNORE` does not actually de-duplicate links when `run_id` is null, because SQLite unique indexes allow multiple NULL values. Old/direct clients can call `log_artifact_use(..., run_id=None)` repeatedly and get duplicate input rows for the same run/artifact/direction. Either make `run_id` non-null before insert (e.g. fall back to `run_name`) or add a second uniqueness strategy for the null-run-id case, otherwise lineage can accumulate duplicates despite the `OR IGNORE`.

abidlabs:
@Saba9 my main concern is this PR seems to reimplement a lot of the existing media/file plumbing for artifact blobs: pending uploads, server upload/download routes, bucket sync, cleanup, and path handling.

Artifacts need their own metadata semantics (versioned names, aliases, manifests, lineage), **but the file/blob lifecycle should probably be shared with media/files as much as possible**. Could we do this:

- a separate artifact Sqlite table with its own metadata: artifacts, versions, aliases, lineage
- a shared asset/blob layer for staging, digest/size validation, pending uploads, server transfer, bucket sync, download, and project cleanup. Some artifact-specific CAS behavior is justified, but I think it'd be better not to duplicate the whole sync stack unless there’s a concrete reason the existing media/files path cannot be generalized.

This should reduce the PR size quite a bit

Saba9:
@abidlabs thanks for the review! I'll try and address the issues and see if I can simplify it as you've suggested.

Saba9:
> @Saba9 my main concern is this PR seems to reimplement a lot of the existing media/file plumbing for artifact blobs: pending uploads, server upload/download routes, bucket sync, cleanup, and path handling.
> 
> Artifacts need their own metadata semantics (versioned names, aliases, manifests, lineage), **but the file/blob lifecycle should probably be shared with media/files as much as possible**. Could we do this:
> 
> * a separate artifact Sqlite table with its own metadata: artifacts, versions, aliases, lineage
> * a shared asset/blob layer for staging, digest/size validation, pending uploads, server transfer, bucket sync, download, and project cleanup. Some artifact-specific CAS behavior is justified, but I think it'd be better not to duplicate the whole sync stack unless there’s a concrete reason the existing media/files path cannot be generalized.
> 
> This should reduce the PR size quite a bit

@abidlabs Thanks, that makes sense. I've moved things a bit further in that direction.

On the blob lifecycle, media and artifact blobs now share the same pending-uploads queue (just tagged with a `kind`), the same replay/drain, the shared temp-file upload handler, and the bucket batching helpers.

I did keep validation and download separate, since that's where the semantics actually diverge. Artifact blobs are content-addressed and digest-checked with versioned downloads, while media/files are just path-addressed UI assets. Given that, I don't think there's a big size win left without weakening the artifact-side guarantees, but if you see a way to generalize that boundary, or think the media path is worth moving onto CAS too, I'm happy to take it further (or split it into a follow-up).

abidlabs:
Sounds good, I'll take another pass, will push tweaks to this PR directly if that's ok?

Saba9:
> Sounds good, I'll take another pass, will push tweaks to this PR directly if that's ok?

That works for me!

abidlabs (approved):
LGTM amazing PR @Saba9! I just made a few tweaks to examples/tests (removed many of the niche unit tests and added a few e2e tests that run on Space to test the complete workflow). Looking forward to the next PR surfacing this in the UI 🔥

abidlabs:
Will go ahead and merge this in!

PR #590, opened 2026-06-29
Use TRACKIO_WRITE_TOKEN for server write token

### LINKED ISSUE #588: feature request: provide preauthenticated write token
In order to point an ML run at a self-hosted instance of trackio, the runner must be authenticated to write to the server, managed by a "write token".
Currently there is no way to pre-authenticate or otherwise externally manage these write tokens.
Instead, a random write token is generated at startup of the trackio server and has to be parsed and passed to the runner every time.

A possible implementation could be this:
```
diff --git a/trackio/server.py b/trackio/server.py
index 4c26e89..f918be9 100644
--- a/trackio/server.py
+++ b/trackio/server.py
@@ -129,7 +129,7 @@ def _flush_loop() -> None:
         _write_queue.clear()
 
 
-write_token = secrets.token_urlsafe(32)
+write_token = os.environ.get("TRACKIO_WRITE_TOKEN", secrets.token_urlsafe(32))
 
 OAUTH_CALLBACK_PATH = "/login/callback"
 OAUTH_START_PATH = "/oauth/hf/start"
```

abidlabs:
We had an existing `TRACKIO_WRITE_TOKEN` env variable, which would get passed by the client. Now, if this is set on the Trackio server, it is used as the write token as well. Closes #588.

### DIFF
--- a/.changeset/angry-onions-take.md
+++ b/.changeset/angry-onions-take.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Use TRACKIO_WRITE_TOKEN for server write token
--- a/.changeset/angry-onions-take.md
+++ b/.changeset/angry-onions-take.md
@@ -1,5 +1,5 @@
 ---
-"trackio": minor
+"trackio": patch
 ---
 
 feat:Use TRACKIO_WRITE_TOKEN for server write token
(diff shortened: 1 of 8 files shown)


abidlabs:
Going to merge this in when green as this is quite a small change, and would like to do a patch release soon.

Saba9 (approved):
LGTM!

abidlabs:
Thanks so much for the review @Saba9!

PR #600, opened 2026-07-02
Add wandb importer

manueldeprada:
## Short description

Adds a local wandb importer feature. Why? The CSV importer gave me issues with my multi-run project: I needed to correct the x-axis name from wandb, was missing the run config, overall not working great.

But warning: the feature was heavily vibe coded. I made sure to supervise, add docs, integration test, etc. It was useful for migrating my runs, but it's totally fine if it's not worth reviewing; I understand these vibecoded PRs can eat too much maintainer time.

## Full description

Adds `trackio.import_wandb("entity/project", project="my-project")` alongside `import_csv` and `import_tf_events`. Each wandb run is imported as a separate trackio run with its full metric history (via `Api.runs()` + `scan_history()`) and config. Same structure as the existing importers: optional dependency behind a `wandb` extra, project-must-not-exist guard, same Space upload tail.

The non-obvious part is custom x-axes. wandb merges everything logged at the same internal step into one history row, so a row can hold a training metric at `_step=5360` *and* an eval metric whose real x is `eval_iter=5000`. A naive import puts both at one step and silently misplaces the eval series. `import_wandb` reads the `wandb.define_metric(..., step_metric=...)` definitions stored in the run config (`_wandb.value.m`, records keyed by `MetricRecord` protobuf field number) and splits each row so every metric lands at its own axis value. Both stored forms are handled: the client writes the step metric name (field `"4"`), the server rewrites it to an index (field `"5"`).

Since those stored definitions don't always survive (they can be lost when eval jobs resume a run), there's a `step_metrics={"eval/*": "eval_iter"}` argument to extend or override them.

### Tests: 
- two mocked-transport tests for the routing and the override
- integration test against the installed wandb client. It asserts the `MetricRecord` field numbers the decoder relies on, creates a real offline run with `define_metric()`, reads back the metric records the client actually wrote, and imports through them, so a wandb update that changes the encoding fails CI rather than corrupting imports. The CI workflow now installs the `wandb` extra (same as `tensorboard`); without it the test skips.

Verified on a real project (15 runs, ~40k history rows, two custom axes): all series land on the correct steps with values matching the wandb UI.

Docs: autodoc entry in `api.md` plus an "Importing Existing Runs" section in `migration.md`.

## AI Disclosure

- [x] I used AI to code, write docs, and test. Supervised myself and edited the end result.
- [ ] I did not use AI

----

## Type of Change

- [x] Documentation update
- [x] Test improvements

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
 
     - name: Install Python dependencies
       run: |
-        uv pip install --system -e '.[dev,tensorboard,spaces]' --prerelease=allow --python "${{ steps.setup-python.outputs.python-path }}"
+        uv pip install --system -e '.[dev,tensorboard,wandb,spaces]' --prerelease=allow --python "${{ steps.setup-python.outputs.python-path }}"
         uv pip install --system pytest --prerelease=allow --python "${{ steps.setup-python.outputs.python-path }}"
 
     - name: Build frontend
--- a/docs/source/api.md
+++ b/docs/source/api.md
@@ -56,6 +56,10 @@
 
 [[autodoc]] import_tf_events
 
+## import_wandb
+
+[[autodoc]] import_wandb
+
 ## save
 
 [[autodoc]] save
--- a/docs/source/migration.md
+++ b/docs/source/migration.md
@@ -82,6 +82,25 @@ wandb.log({"training_results": table})
 wandb.finish()
 ```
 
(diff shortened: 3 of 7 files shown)


abidlabs:
Hi @manueldeprada this is a bit out-of-scope for Trackio currently. If you open up a smaller-scoped PR to fix import_csv to handle the issue that you described, that would be more reviewable.

PR #602, opened 2026-07-04
feat: add Artifact.add_reference to reference external objects without copy

adrien-grl:
## Short description

Add artifact references (`add_reference`): catalog external data without copying

### Motivation

Trackio's artifact tracking (#586) versions **file bundles**: every file is hashed and copied into local content-addressed storage (CAS), then synced. That's right for models and small datasets, but unusable for **large or already-durable data** — a temporary multi-TB archive living on local disk, or objects already in an object store — which should not be duplicated.

This PR adds **reference entries**: manifest entries that record a URI (+ size + checksum) **without staging any blob**, mirroring `wandb.Artifact.add_reference`. You get versioning, dedup, lineage, and aliases over data that never moves — cataloging, not copying.

It completes the `add_reference` capability from the original feature request (#546), which #586 implemented for files only.

- Original issue: #546
- Artifact tracking (files): #586

### What this adds

```python
artifact = trackio.Artifact(name="training-set", type="dataset")
artifact.add_reference("file:///mnt/nvme/datasets/corpus.parquet")   # sha256 + size, no copy
artifact.add_reference("s3://my-bucket/shards/")                     # a prefix — expands to one entry per object
artifact.add_reference("gs://my-bucket/eval.json")
artifact.add_reference("hf://datasets/username/my-dataset/train.parquet")
trackio.log_artifact(artifact)
```

```python
add_reference(uri, name=None, checksum=True, max_objects=None, size=None, digest=None)
```

A pluggable per-scheme handler resolves size + a checksum, expands a directory/prefix into one entry per object, and (on `download()`) fetches the bytes:

| Scheme | Resolve (size + checksum) | Expand dir/prefix | Fetch on `download()` | Dependency |
|---|---|---|---|---|
| `file://` | size + sha256 `digest` | ✅ recursive | copy | stdlib |
| `http(s)://` | `Content-Length` + `ETag` | — | stream GET | bundled `httpx` |
| `hf://` | size + LFS sha256 or git blob id | ✅ | `HfFileSystem.get_file` | core `huggingface_hub` |
| `s3://` | size + `ETag` | ✅ | `download_fileobj` | optional `trackio[s3]` |
| `gs://` | size + `ETag` | ✅ | `download_to_filename` | optional `trackio[gcs]` |
| Azure blob¹ | size + `ETag` | ✅ | `download_blob` | optional `trackio[azure]` |
| anything else | — (opaque pointer, warns) | — | raises | none |

¹ `https://<account>.blob.core.windows.net/<container>/<blob>` (matched by netloc, as wandb does).

`references` / `get_entry_uri()` expose the recorded URIs without downloading.

A local `file://` URI must exist when added — a missing one **raises** (matching wandb's `LocalFileHandler`; a missing local file is almost always a typo). Remote references, by contrast, are recorded even when currently unreachable (you may be cataloging data behind auth or not-yet-uploaded), failing only at `download()`.

#### Module layout

Reference handling is done a new module:

- **`references.py`** — everything about references: the primitives (`validate_reference_uri`, `local_path_from_file_uri`, `default_reference_name`, and the `is_reference_entry` manifest discriminator) plus the I/O layer — an abstract `ReferenceHandler` (`matches`/`resolve`/`fetch`/`hint`) with one subclass per scheme (`File`/`Http`/`Azure`/`S3`/`Gcs`/`Hf`/`Tracking`), a `_select` dispatcher, and a `TrackingHandler` fallback. The optional cloud SDKs are imported lazily inside the handler that needs them, so `import trackio` pulls in none of them (only `httpx`, already a core dependency).

This mirrors how wandb's client is organized. `wandb` keeps content-addressed storage (`artifact_file_cache.py`, `staging.py`, `storage_layout.py`) separate from its per-scheme storage handlers: an abstract `StorageHandler` (`can_handle`/`store_path`/`load_path`) subclassed once per scheme in a `storage_handlers/` package, selected by a `MultiHandler` and defaulting to a `TrackingHandler`. `reference.py` follows the same design — an abstract `ReferenceHandler` (`matches`/`resolve`/`fetch`/`hint`) with one subclass per scheme (`File`/`Http`/`Azure`/`S3`/`Gcs`/`Hf`/`Tracking`), a `_select` dispatcher, and a `TrackingHandler` fallback. All these primitives live in a single file at the moment, but could be split into multiple files in a folder at a later step.

### Manifest format & backward compatibility

`ManifestEntry` is extended **additively**: `path`/`digest`/`size` required (exactly the original file entry); optional `ref` (URI). An entry is a reference iff it carries `ref`.

- **File entry** (unchanged): `{"path", "digest", "size"}`
- **Reference entry**: `{"path", "ref", "digest", "size"}`

`digest` is a single field holding whatever checksum the entry has — a real sha256 for files, `file://`, and LFS `hf://`; the provider's opaque **ETag** for cloud/HTTP references; and, **when no checksum can be obtained (no ETag, a missing cloud SDK, an opaque scheme), the URI itself** — always with a warning naming the fix. This mirrors `wandb`, whose entry `digest` likewise holds an md5 for files, an ETag for cloud refs, and the URL when nothing better exists. A *file* entry's `digest` stays a validated sha256 (it is the CAS key); a *reference's* `digest` is an opaque token and is not sha256-validated. 

`SQLiteStorage._canonical_manifest` branches on `ref` and emits the **exact previous shape** for file entries, so **every existing artifact's `manifest_digest` is byte-for-byte unchanged** (locked by a test). Reference identity folds URI + size (+ `digest`), so dedup and versioning fall out of the existing `UNIQUE(artifact_id, manifest_digest)` constraint. No SQLite migration (manifest is JSON TEXT; additive keys suffice).

#### No hard dependencies; optional SDKs lazily imported

`http(s)://` and `hf://` add nothing (`httpx` and `huggingface_hub` are already a dependency). Cloud SDKs are optional extras (`trackio[s3]`/`[gcs]`/`[azure]`) imported **only when their scheme is used with checksumming** (we also mirrored what `wandb` does here). When a needed client is absent, resolve **degrades to an un-checksummed record** (with a one-line "install trackio[…]" warning); `download()` **raises** because the caller has explicitly asked for the bytes which cannot be reached.

#### `download()` fetches references

Unlike files (copied into CAS at log time), reference bytes stay put until you fetch them, and users need to call `download()` to fetch them. This is the expected behavior when a caller explicitly asks to materialize an artifact. Because that can transfer a lot of data, `references` / `get_entry_uri()` does not trigger any download. `download()` is atomic and idempotent.

#### Directory / prefix expansion + `max_objects`

A `file://` directory or an `s3://`/`gs://`/`hf://`/Azure prefix expands into one entry per object, each named by its path relative to the prefix (nested under `name=`). This is what makes referencing a whole dataset ergonomic (and is why wandb has `max_objects`). `max_objects` guards the fan-out — listing/checksumming millions of objects, and a manifest bloated to millions of entries — and **raises** if exceeded. The relative name is computed against the prefix's directory boundary, so a bucket-root reference keeps each object's full key and a prefix like `logs` never mangles a sibling `logs.txt`; object keys are percent-encoded in the stored `ref` (and decoded on fetch) so keys containing spaces round-trip through validation. `checksum=False` skips only the per-object checksum — an object-store prefix still expands.

#### Permissive scheme handling

`validate_reference_uri` accepts any URI (rejecting only empty/schemeless/whitespace/NUL URIs). A URI whose scheme has no dedicated handler is recorded as an opaque pointer with a warning, and raises if you try to `download()` it — wandb's behavior. This lets users reference `dvc://`, custom schemes, etc., before trackio grows a handler for them.

#### No blob staging or sync for references

References stage nothing into CAS, enqueue nothing in `pending_uploads`, sync nothing to Spaces/buckets, and never crash a manifest-iterating path on a missing `digest`. An LLM-made audit of every manifest-touching site found only two needing guards — `Run.log_artifact` (remote blob upload) and `server.artifact_log` (validation); every other sync path walks the on-disk CAS directory and excludes references. `/artifact_blob/…` returns a clean 404 for a reference digest (tested against a real 200 for a staged file).

### wandb `add_reference` parity

| Aspect | wandb | This PR |
|---|---|---|
| Auto-detect size + checksum (no manual `size`) | ✅ | ✅ |
| `file://` (+ directory expansion) | ✅ | ✅ |
| `http(s)://` | ✅ (`requests`) | ✅ (`httpx`) |
| `s3://` / `gs://` / Azure | ✅ (lazy SDKs) | ✅ (same) |
| Hugging Face (`hf://`) | ❌ | ✅ (core `huggingface_hub`) |
| Prefix/dir expansion + `max_objects` | ✅ | ✅ |
| `download()` fetches references | ✅ | ✅ |
| Permissive fallback for unknown schemes | ✅ | ✅ |
| Missing-SDK behavior | ⚠️ raises at add | degrades at add, raises at download |
| Bytes never copied into the store | ✅ | ✅ |
| Auto-fetch checksum *verification* on download | ✅ | ✅ |
| `wandb-artifact://` | ✅ | ❌ (out of scope) |

### Tests

New `tests/unit/test_artifact_references.py` (72 tests). Cloud SDKs are mocked at the handler seam; `file://` runs for real and `http(s)://` via a mocked `httpx`.

### Notes

- **No new hard dependencies.** New optional extras: `trackio[s3]` (boto3), `trackio[gcs]` (google-cloud-storage), `trackio[azure]` (azure-storage-blob + azure-identity). `hf://`/`http(s)://` add nothing.
- **No existing `manifest_digest` values change.**
- **No UI changes** (the CLI `get artifact` view shows a reference's URI; the dashboard doesn't render artifacts yet).
- **`.gitignore`:** added `/artifacts/`. `Artifact.download()` (from #586) defaults to writing under `./artifacts/`, so running the artifact test suite from the repo root leaves an untracked `artifacts/` dir; this ignores it.

#### Two calibration choices worth a reviewer's eye

- **`max_objects` default = 10,000** (wandb defaults to ~10M). trackio stores the whole manifest as a single JSON blob in SQLite and re-serializes it on every log, so a very large expansion is costlier here; a lower, override-able default felt safer. Easy to change.
- The digest's type is currently called `Sha256Digest`: with this PR, the digest is not necessarily sha256 anymore. For instance, an ETag is not sha256 and is still considered a valid digest by this PR. It might be better to rename this type, but the blast radius is large so we scoped it out of this PR without explicit maintainer approval.

## AI Disclosure

We encourage the use of AI tooling in creating PRs, but any non-trivial use of AI needs be disclosed. E.g. if you used Claude to write a first draft, you should mention that. Trivial tab-completion doesn't need to be disclosed. **You should self-review all PRs, especially if they were generated with AI**. 

-----

- [X] I used Claude Opus 4.8 to write most of the code, but I thoroughly reviewed and improved it afterwards.
- [ ] I did not use AI

----

## Type of Change

- [ ] Bug fix
- [X] New feature (non-breaking)
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

If this PR closes an issue, please link it below: #546

Closes: 

## Testing and linting

   ```bash
   python -m pytest
   ```

   ```bash
   ```


abidlabs:
Thanks @adrien-grl! cc @Saba9  for visibility

Saba9:
Thanks for the PR @adrien-grl! Taking a look.

Saba9:
@adrien-grl this looks great. A few things I'd like to sort before we merge, and I think most of them come down to one scope call.

**1. Can we drop `size=` and `digest=` from `add_reference`?**
wandb's version doesn't take either. The bigger reason is that the caller `digest=` causes the false positive in point 2: we validate it as a real sha256 (`artifact.py:367`), then on download compare it against a freshly-probed ETag, so it can never match. On top of that the `digest` field is already overloaded three ways: a sha256 for `file://`/LFS, the provider ETag for cloud/HTTP, and the URI itself when nothing can be probed (`artifacts.md:133`). Letting callers also pass a sha256 makes that worse. Drop it and the kind just follows from the scheme.

While we're in there: the field is typed `Sha256Digest` everywhere even though ETags and URIs flow through it too. A `Sha256Digest | ETag | URIStr` union on the reference paths would be more honest, and since these are `NewType`s it doesn't touch the manifest.

**2. The drift check should warn, not raise** (`artifact.py:94-99`).
Download re-probes the source and raises "has changed since it was logged" on any mismatch. Two ways that fires on unchanged content: recorded sha256 vs probed ETag (which goes away once `digest=` is gone), and ETag vs ETag, since ETags shift with multipart chunk size, CDN node, gzip, and so on. A drifted ETag shouldn't block a download that's otherwise fine. Related: `artifacts.md:141` says every object "is verified against its recorded checksum," but we skip a file that's already on disk (`:84`), so that line needs updating too.

**3. Signed URLs leak into synced manifests.** 
If someone passes a presigned `https://` or Azure SAS URL, we store it verbatim as `ref` (and as `digest` when the source can't be probed, `artifact.py:395`), and manifests sync to a Dataset that might be public, so the credential ends up published. A `warnings.warn` in `add_reference` when the query looks signed (`X-Amz-Signature`, `X-Goog-Signature`, `sig`, `token`) would catch it before we persist anything. `s3://` and `gs://` URIs don't carry credentials, so they're fine as-is. wandb doesn't warn on this, so it's slightly beyond parity, but it's cheap.

One more, separate from the above: logging a reference against a Space that predates this PR throws a bare digest-validation error that looks like a user mistake. Worse, a `file://` reference (which has a real sha256) can get accepted by an old server as a normal file entry whose blob was never uploaded. A clearer client-side message would cover both. Fine as a follow-up.

Thanks again, really nice work here.

adrien-grl:
@Saba9 Thank you very much for your review. I addressed your comments. Can you please have another look and tell me if there's anything else I should change?
Thank you.

Saba9 (approved):
Looks good to me!

abidlabs:
My issue with this is that this adds a lot of vendor-specific logic in `trackio/references.py` that we now have to maintain. I don't think we should do that. IMO we just keep the `ResolvedReference` abstract class and then document how to use it with other vendors.

abidlabs:
I'll adapt this PR to do that instead (we can include some documentation showing how to use it with other vendors) but we shouldn't include that in our codebase.

abidlabs (approved):
LGTM merging!

PR #603, opened 2026-07-06
Move artifact downloads under .trackio

abidlabs:
## Summary
- move default artifact downloads from ./artifacts to ./.trackio/artifact-downloads
- update artifact docs for the new default path
- add a unit assertion that the default download path stays out of ./artifacts

## Tests
- python -m pytest tests/unit/test_artifact_e2e.py
- python -m ruff check trackio/artifact.py tests/unit/test_artifact_e2e.py

### DIFF
--- a/.changeset/artifact-download-location.md
+++ b/.changeset/artifact-download-location.md
@@ -1,0 +1,5 @@
+---
+"trackio": patch
+---
+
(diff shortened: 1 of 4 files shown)


abidlabs:
This cleans up git noise, will merge it in.

PR #604, opened 2026-07-06
docs: clarify artifact types

abidlabs:
Add an Artifact types section to the artifacts docs. It explains that Trackio artifact `type` is free-form and documents common conventions such as model, dataset, evaluation, report, and unspecified.


abidlabs:
Just a quick docs PR, merging

PR #605, opened 2026-07-06
Panels-per-row and row pagination controls to panel groups

manueldeprada:
## Short description

Metric groups currently pack every panel into one flex row that wraps based on a fixed 350px minimum width. With 4 panels this looks fine; with 6+ panels each chart gets squeezed too narrow to read, and there was no way to control it.

This adds two small controls to the top-right of each group's header (next to the title and fold arrow):

- Panels per row: a dropdown to pick how many panels fit per row (capped; a 3-panel group won't offer "4/row").
- Rows shown: a "Rows X-Y of Z" control with ‹/› navigation between pages of rows.

Screenshots:
image
image

Both settings are per-group and default to 4/row, all rows shown. So nothing changes visually until you touch the controls.

## AI Disclosure

- [x] I used AI to code the feature.
- [ ] I did not use AI

----

## Type of Change

## Related Issues

If this PR closes an issue, please link it below: 

Closes: 

## Testing and linting

   ```bash
   python -m pytest
   ```

   ```bash
   ```

### DIFF
--- a/tests/ui/test_metrics_group_pagination.py
+++ b/tests/ui/test_metrics_group_pagination.py
@@ -1,0 +1,128 @@
+from playwright.sync_api import expect, sync_playwright
+
+import trackio
+
+
+def _log_grouped_metrics(project):
+    trackio.init(project=project, name="run_0")
+    for s in range(5):
+        metrics = {f"big/m{i}": float(s + i) for i in range(8)}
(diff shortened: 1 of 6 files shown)


abidlabs:
Hmm I prefer the minimalist UI without these dropdowns in the corner. We could potentially add a dropdown for # of plots per row in the sidebar under the smoothing factor -- could you try creating a new PR with that?

PR #611, opened 2026-07-09
Logbook: auto-capture output artifacts and live-embedded dashboard cells

abidlabs:
## What

Three logbook improvements aimed at making agent-run experiments richer and more engaging to watch:

### 1. Auto-capture output files as artifact cells
After `trackio logbook run -- <command>` finishes, files the command **created or modified** with common ML model/data extensions (`.pt`, `.safetensors`, `.ckpt`, `.parquet`, `.csv`, `.jsonl`, …) are recorded as **path-reference artifact cells** right after the run cell — path, size, and inferred type only (no copy, no artifact-store commit, not pushed to a Bucket on publish). Capped at the 10 largest files with an overflow note; skips hidden dirs, venvs, and empty files. Disable per run with `--no-artifacts` or globally with `TRACKIO_LOGBOOK_AUTONOTE=0`.

### 2. Embedded `dashboard` cells, live at `trackio.init()`
New first-class `dashboard` cell type that **fully embeds** a project's Trackio dashboard in the page body. `trackio.init()` inside a logbook workspace adds one immediately (deduped per project per page), so anyone watching the logbook preview sees training metrics stream in real time — the preview live-reloads within ~1.5s and the embedded dashboard updates itself.
- `logbook serve` now lazily launches the local dashboard in-process and advertises its (tokenless) URL at `/dashboard-info.json`; the viewer iframes it. Published logbooks keep working via the existing sentinel → Space promotion.
- `trackio.finish()` no longer writes a run cell (superseded by the dashboard cell).
- Manual CLI: `trackio logbook cell dashboard <project> [--space owner/name]`.

### 3. `trackio skills add` central install
Bare `trackio skills add` now installs to the central `.agents/skills` location (like `hf skills add`); agent flags remain optional for symlinks/commands/hooks.

## Testing
- 17 new unit tests in `tests/unit/test_logbook.py` (233 total passing; only pre-existing no-CUDA GPU tests fail locally).
- Added an autouse conftest fixture disabling logbook autonote so the new init hook can't write into a developer's real logbook during tests.
- Verified live end-to-end: mid-training screenshot shows the embedded dashboard streaming loss/acc inside the logbook preview; `--no-artifacts`, dedupe, publish rewrite, and `/dashboard-info.json` (single launch, no token leak) all exercised.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #597, opened 2026-07-01
Add experiment logbooks: shareable static-Space lab notebooks

abidlabs:
## Motivation

Most ML experimentation is still captured in fragments: terminal scrollback, one-off notes, dashboard links, checkpoint paths, figures, raw data files, and agent summaries. The valuable output of an experiment campaign is not only the final metric, but it is the trace of what was tried, what changed, what failed, what artifacts were produced, and why the next decision made sense.

This PR adds **Trackio logbooks**: local-first, Hub-native lab notebooks for experiment campaigns. A logbook is just Markdown files in `./.trackio/logbook/` that you (or your agent) append to as the campaign unfolds. It renders as a static Hugging Face Space for humans — with a resources sidebar collecting every model, dataset, paper, job, and artifact mentioned on a page, and exposes a compact, token-efficient view for agents through `trackio logbook read`.

https://github.com/user-attachments/assets/6e30dd4d-657d-4ea4-b2ec-47f5fae3633b

## Starting a logbook

From the terminal:

```bash
trackio logbook open              # scaffold ./.trackio/logbook/ in the current repo
trackio logbook serve             # preview locally
trackio logbook publish user/name # publish as a static HF Space (auto-syncs afterwards)
```

The easiest way, though, is through your coding agent. `trackio skills add --claude` (or `--cursor`, `--codex`, `--opencode`, `--pi`) installs the Trackio skill **plus a `/logbook` slash command**. Typing `/logbook` opens a logbook in the current directory and puts the agent in logbook mode for the session: it adds a page per experiment, logs findings as it works, runs commands through `trackio logbook run` so commands/scripts/outputs are captured verbatim, and keeps the index page a clean table of contents. `/logbook --cell "..."` records a specific finding, and `/logbook --publish user/name` publishes the Space. (For Claude Code, it also installs a hook that mirrors the agent's todo list into logbook pages.)

Trackio itself auto-captures into an open logbook: `trackio.finish()` records each run with a link to its dashboard, and `trackio.log_artifact()` records an artifact cell — no manual logging needed for those.

## Cell types

Everything is stored as plain Markdown with typed cells, so pages can always be edited directly.

**Markdown cells** — findings, decisions, and context. URLs written in the body are collected into the page's resources sidebar (models, datasets, Spaces, papers, repos, jobs, buckets), with inline chips that cross-highlight their sidebar entries on hover. Fenced code blocks render with syntax highlighting.

```bash
[...]

PR #614, opened 2026-07-10
Render logbooks as single-page documents and improve UI

abidlabs:
https://github.com/user-attachments/assets/b4916446-2e6f-43ac-ae55-90b9ab032e14

This builds on the viewer introduced in #597 and preserves the live-preview and dashboard behavior added in #609 and #611.


abidlabs:
Going to merge this in so that we can test the logbook workflow end to end

Issue #607, opened 2026-07-07
[Proposal] Artifact registry: gather artifacts across projects


adrien-grl:
# Why

My team is migrating off W&B following their recent ToS changes. We were convinced by Trackio's approach and we would like to fully migrate. Artifacts support (#586, #601, and hopefully references via #602) already covers most of what we used. The last missing piece for a full migration is the **artifact registry**. We're prepared to build it in a fork if it's out of scope for Trackio, but we'd much rather design it with you and contribute it upstream.

# What an artifact registry is

A registry is a curated, cross-project catalog of artifact versions, organized into **collections**. Where a project's artifact list answers "what did my experiments produce?", a registry answers "what is the model running in production right now?"

A collection represents one asset as a product — e.g. the churn model, the golden eval set — and aggregates the versions of it worth publishing, regardless of which project or retraining produced them. Versions are **linked** into a collection (pointers, no copy) and promoted through lifecycle stages by moving aliases like `staging` and `production`.

The primitive is generic: the same mechanism gives you a model registry, a dataset registry, a prompt registry — any curated family of artifacts.

# Why it's nice

- **One source of truth for consumers.** Deployment, evaluation, and CI resolve `model:production` without knowing anything about experiment projects. Promoting a new version breaks nothing downstream — the alias moves, consumers don't change.

- **The research → production hand-off point.** Training publishes into the registry; everything after consumes from it. This is the boundary between "tracker" and "system of record".

- **Auditability.** Which version is deployed, since when, what it replaced, and what run produced it — connected to the lineage that #586 / #601 already track.

- **Automation hook.** "A version was promoted to production" is the natural event for CI/CD to react to (webhook or polling; mechanism TBD, but the registry is where the event lives).

# Features we need

- Create registries and typed collections within them (a collection accepts one artifact type).
- Link an existing artifact version into a collection (pointer, no copy).
- Collection-side version numbering (`v0`, `v1`, ... per collection), since linked versions come from different source artifacts whose own numbering neither aligns nor stays unique.
- Alias-based promotion, one version per alias per collection, same semantics as the existing per-artifact aliases.
- Resolution and lineage through the registry: consuming `model:production` records the run ← registry ← source-artifact chain.
- An audit history of link/promote operations.
- CLI support (and UI eventually, as a follow-up in the spirit of #586 → #601).

# API sketch

```python
# Publish: link an already-logged artifact version into a collection
run.link_artifact(artifact, "models/churn-model", aliases=["staging"])

[...]

PR #645, opened 2026-07-22
Add artifact registry: storage and local publishing API

adrien-grl:
## Short description

How projects, artifact versions, collections, links, and aliases relate

This is the first PR of the artifact registry proposed in #607: the storage layer and the Python publishing API, for **local and bucket-backed registries**.
In order to minimize the size of the PRs and to enable discussions, the task of implementing the artifact registry is split into multiple PR. Current plan:
- **PR 1:** storage and publishing API — local, plus bucket-backed registries (this PR)
- **PR 2:** registry resolution (`use_artifact` / `download` through a registry location)
- **PR 3:** CLI
- **PR 4:** UI

A registry is a curated, cross-project catalog. It holds typed **collections**, into which you **link** existing artifact versions (links are pointers, no copy) and promote them by moving **aliases** like `staging` and `production`:

```python
registry = trackio.Api().create_registry("models", description="Models we deploy")

run = trackio.init(project="my-experiments")

# Create a model here
...

artifact = trackio.log_artifact("model.pt", name="resnet", type="model")
linked = run.link_artifact(artifact, "registry-models/my-model", aliases=["staging"])

linked.version, linked.aliases                # "v0", ["latest", "staging"]  (collection-side)
linked.is_link, linked.source_qualified_name  # True, "churn-experiments/resnet:v0"

registry.collection("churn-model").links      # the linked versions, newest first
registry.events()                             # append-only audit log

# Artifact.link / Artifact.unlink work too (the artifact must be logged):
linked = artifact.link("registry-models/churn-model", aliases=["staging"])
linked.unlink()
```

## Remote registries: a registry is a bucket

A local registry only exists on the machine that wrote it, which left a run that logs to a Space with nowhere to publish. Passing `bucket_id` (or setting `TRACKIO_REGISTRY_BUCKET_ID`) backs the registry with a Hugging Face bucket instead:

```python
registry = trackio.Api().create_registry("models", bucket_id="my-org/models-registry")

run = trackio.init(project="my-experiments", space_id="my-org/my-dashboard")
artifact = trackio.log_artifact("model.pt", name="resnet", type="model")
run.link_artifact(artifact, "registry-models/my-model", aliases=["staging"],
                  bucket_id="my-org/models-registry")
```

Every writer talks to the bucket directly with its own HF credentials — no Space and no server in the write path — so a registry is org-wide and outlives any single Space. `Api().registry`, `Artifact.link` and `Artifact.unlink` take the same argument, and a linked artifact remembers its bucket so `linked.unlink()` returns to the right registry.

**Why the local layer changed.** Object storage offers no compare-and-swap, so two machines cannot safely read-modify-write a shared `.db` (or a shared state file) — the second writer silently overwrites the first. The only safe primitive is adding objects under unique keys, so registry mutations are now event-sourced: each mutation is built as an event, folded into the projection tables by one reducer, and appended to the log. Locally that is the same single transaction under the same process lock as before (all existing tests pass unchanged); in a bucket each event is one immutable object and readers fold the log into a local projection cache.

```
trackio/registries/<registry>/registry.json            name, description, created_at
trackio/registries/<registry>/events/<event_uid>.json  one object per mutation
```

Two consequences, both deliberate and documented: **collection versions are assigned by the fold**, not claimed by a writer (so concurrent links get distinct, never-reused numbers ordered by `event_uid` = timestamp, writer, sequence), and **alias moves, `latest` included, are last-writer-wins** under that same order. Links stay pure pointers — nothing is copied into the registry bucket. They do now record the source's storage coordinates (`source_space_id` / `source_bucket_id`, `NULL` when local), because once a registry is shared, *where* the bytes live can no longer be derived from the source coordinates alone. No content digest is snapshotted on the link (keeping 44f1ab8's decision). Copying (pinning) a version's bytes into the registry, and resolution through a registry location, remain follow-ups.

## Design decisions

Of course, any of these decisions are open for discussion. Consider this PR as a first draft.

- **Storage layout.** A registry is a regular trackio project under the reserved name prefix `registry-` (registry `models` → `registry-models.db`), reusing the per-project SQLite / locking / connection machinery. The registry tables live alongside the standard project schema, so project-listing and dashboard queries keep working unchanged. `validate_project_name` rejects the `registry-` prefix, so a run can't be logged into a registry DB by accident.
- **New modules.** `registry_storage.py` and `registry.py`, rather than growing `sqlite_storage.py` (already ~5k lines). They build on `SQLiteStorage`'s connection/lock primitives and mirror the existing `artifact.py` / `sqlite_storage.py` split.
- **Monotonic collection versions.** The one deviation from the sketch schema: `collections.next_version`, a per-collection counter assigned under the existing per-project process lock (race-safe, same pattern as artifact versioning). Unlike `MAX(version)+1`, the counter never reuses a number after an unlink, so a `collection:vN` that was ever published can't silently come to mean something else.
- **Explicit registry creation.** `Api().create_registry(name, description=None)` and `Api().registry(name)` mirror W&B. Registries are never created implicitly and linking or creating a collection in a missing registry raises. Collections still auto-create on first link (also W&B behavior).
- **Alias semantics** follow per-artifact aliases: upsert = promotion (one version per alias per collection), and an explicit promotion may move an alias backward (rollback). `vN`-shaped aliases are rejected. `latest` auto-follows the newest link; passing `"latest"` in `aliases` is a no-op, not an error (as W&B does).
- **Linked artifacts carry their provenance.** `is_link`, `source_name`, `source_version`, `source_project`, `source_qualified_name` — equal to the plain properties on non-links, mirroring W&B's `source_*` family. Linking an already-linked artifact links its *source* version directly, so links never chain. `download()` through a registry location raises an explicit `NotImplementedError` until resolution lands (see below).
- **Two ways to link, one path.** `Run.link_artifact(artifact, target, aliases)` (auto-logs a draft, records the publishing run) and `Artifact.link(target, aliases)` (requires an already-logged artifact, records no run) both funnel through one internal `_link_version`. `Artifact.unlink()` removes a link. All three are local for now.
- **Audit log.** `registry_events` (append-only) records one event per mutation — registry/collection `create`, `link`, `promote`, description `update`, and `unlink` — in the same transaction as the mutation. Link and promote events record the publishing run (`run_name` / `run_id`); promote events also record the `previous_version` the alias moved from.

## W&B parity

- **Signature-compatible.** `Run.link_artifact(artifact, target_path, aliases=None) -> Artifact` matches W&B's method, parameter names included, so keyword callers port unchanged. As in W&B, the returned artifact sits at its registry location (`version` / `aliases` are the collection-side ones), and `latest` is applied automatically to the most recent linked version. W&B has no module-level `wandb.link_artifact` (only `run.link_artifact`); this mirrors that.
- **Auto-logging.** Like W&B, `link_artifact` first logs a not-yet-logged artifact to the run's project, then links it.
- **Instance-only.** `link_artifact` takes an `Artifact` instance; a string raises a `TypeError` with guidance. Linking publishes one specific version, and the instance you hold *is* that version. The "select then publish" flow is `use_artifact("resnet:v3")` (or an alias) followed by `link_artifact`, same as W&B.
- **Target path.** Structurally identical to W&B — a prefix-reserved project followed by the collection. We use `registry-models/my-model` where W&B uses `wandb-registry-model/my-model`: same layout, shorter prefix. This is a deliberate ergonomic choice; the target string is trackio-specific either way.

Divergences, all deliberate:

The remaining divergences will be resolved by the resolution PR, except for:

- **`create_registry` omits `visibility` / `organization` / `artifact_types`.** Org visibility and RBAC are descoped in #607; the registry-level type allowlist is additive later. `description` is supported.

## Schema (in registry DBs only)

- `collections(id, name UNIQUE, type, description, next_version, created_at)` — one artifact type per collection, enforced at link time.
- `collection_links(id, collection_id, collection_version, source_project, source_artifact, source_version, source_space_id, source_bucket_id, created_at, UNIQUE(collection_id, source triple), UNIQUE(collection_id, collection_version))`.
- `collection_aliases(collection_id, alias, link_id, PRIMARY KEY(collection_id, alias))`.
- `registry_events(id, event_uid UNIQUE, ts, kind, payload)` — `event_uid` is `<compact-utc>-<writer>-<seq>`, which totally orders events across writers and makes replay idempotent.

## Notes for reviewers

- **The `registry-` prefix is a deliberate reuse of the project namespace**, not a parallel store — that's what lets a registry inherit SQLite, locking, and hub-sync for free.
- **The event log is now the authoritative form of a registry**; the tables are a fold of it. That is what makes one implementation serve both a local database and a bucket, and it is the piece worth reviewing most closely (`registry_storage.apply_event_cursor`, `registry_bucket.BucketRegistryStorage`).

## AI Disclosure

-----

- [x] I used AI to... write the first draft of this PR (Claude); I made the design decisions above with it and self-reviewed the full diff.
- [ ] I did not use AI

----

## Type of Change

- [ ] Bug fix
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

Part of: #607 (first PR of the sequence; the issue stays open for the follow-ups)

## Testing and linting

- `pytest --deselect=tests/ui --deselect=tests/e2e-spaces --deselect=tests/unit/test_gpu_hardware.py`: **354 passed, 1 skipped**
- `pytest tests/unit/test_registry.py`: **44 passed**
- `ruff check --fix --select I && ruff format`: clean
- `tests/ui` and `tests/e2e-spaces` untouched by this PR (no frontend or server changes).


abidlabs:
@adrien-grl thanks for opening this up! I'll review this and try to get this merged in next week. I'd suggest actually also doing:

> PR 2: remote (Space / self-hosted) publishing

as part of this PR. 

Thinking about the right API that will cover both local and Space/self-hosting will be important and may change the implementation a bit.

abidlabs:
@adrien-grl are you interested in expanding this to cover artifact registries for Trackio dashboards published to Spaces? If not, I'm happy to adapt this PR

trackio/registry_storage.py
```diff
@@ -0,0 +1,766 @@
...
+        """Create the registry's database and append a ``create`` event.
+
+        The database gets the standard project schema plus the registry
+        tables. An optional `description` is recorded on the create event.
+        Raises ValueError when the registry already exists. Registries are
+        never created implicitly: linking or creating a collection in a
+        registry that does not exist raises.
+        """
+        project = registry_project_name(registry)
+        if RegistryStorage.registry_exists(registry):
```
abidlabs: This existence check happens before any process lock, while schema creation, metadata writes, and the audit event are split across several later transactions. Two concurrent `create_registry("models")` calls can both observe “missing,” then both return success and append separate `create` events; a crash after the DB is initialized also leaves a partial registry that every retry treats as already complete. Please make the existence check plus registry marker/metadata/event creation atomic under one process lock (checking again after the lock is acquired).

trackio/registry_storage.py
```diff
@@ -0,0 +1,766 @@
...
+                resolved = RegistryStorage._resolve_link_cursor(
+                    conn, registry, collection, collection_version
+                )
+                link_id = resolved["link_id"]
+                alias_rows = conn.execute(
+                    "SELECT alias FROM collection_aliases WHERE link_id = ?",
+                    (link_id,),
+                ).fetchall()
+                removed_aliases = sorted(r["alias"] for r in alias_rows)
+                conn.execute(
```
abidlabs: If this link owns `latest`, deleting its aliases and the link leaves surviving older links with no `latest` alias. For example, link v0 and v1, then unlink v1: v0 remains with `aliases == []`, contradicting the public invariant that `latest` always follows the newest linked version. When removing the current latest, please reassign `latest` to the highest remaining `collection_version` in this same transaction (and cover the intermediate state before another link is added).

abidlabs (commented):
Found two registry-state correctness issues.

abidlabs (approved):
LGTM @adrien-grl! Thank you for initiating this work. I added HF Bucket support for remote registries.

The CLI/UI is not included as part of this PR. Lmk if you are interested in taking that on in coming days, otherwise I'll work on it.

adrien-grl:
Hey, thank you for finishing up the work!
Sorry I couldn't attend fast enough, things have been pretty busy on our side lately.
I'd be happy to contribute the CLI/UI part, but I'm afraid I won't be able to attend until early September.
Let me know if that works for you!

abidlabs:
No worries at all, thanks again for driving this work. I'll put a PR together for CLI/UI later this week or next week. If you have time to provide feedback, that'd be welcome

PR #647, opened 2026-07-23
Increase smoothing upper bound to 100

### LINKED ISSUE #646: Increase smoothing to 50
Requested by @adithya-s-k

> Hey just wanted to ask if it will be possible to increase the upper bound on the smoothing option to something like 50 or 100 ?

abidlabs:
Raises the maximum value of the dashboard's smoothing slider from 20 to 100, as requested in #646.

Smoothing is implemented as a centered moving average in `trackio/frontend/src/lib/dataProcessing.js`, where the window size is already clamped to the number of data points (`Math.max(3, Math.min(windowSize, rows.length))`), so values up to 100 are safe even for short runs.

https://github.com/user-attachments/assets/e8c580ed-5731-4e3d-ad9e-a4e8cec33d84

Closes #646

### DIFF
--- a/.changeset/green-rockets-invite.md
+++ b/.changeset/green-rockets-invite.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Increase smoothing upper bound to 100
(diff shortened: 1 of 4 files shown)


Saba9 (approved):
👍

abidlabs:
Sweet! Will get this in and do a release

PR #309, opened 2025-10-14
Add histogram support with wandb-compatible API

abidlabs:
Adds histogram logging functionality matching wandb's Histogram API. Implements the Histogram class that accepts sequences or pre-computed numpy histograms, with configurable bin counts. Includes UI visualization using bar charts and comprehensive test coverage. Closes: #308

image

AI disclosure: this PR was initially implemented with AI, then I self-reviewed it and modified the implementation a bit before opening the PR.

PR #649, opened 2026-07-23
Fix histograms not visible in the dashboard

### LINKED ISSUE #637: [BUG] Histograms no visible in the dashboard
The dashboard does not show histogram metrics (trackio==0.31.5).

# Expected behavior

After using `trackio.log({"logits": trackio.Histogram(x)})`, a histogram chart should appear in the dashboard.

# Actual behavior

The histogram is stored in the sqlite database.

```json
{
    "logits": {
        "_type": "trackio.histogram",
        "bins": [ 7.988692283630371, "...", 12.998191833496094 ],
        "values": [ 1, "...", 1 ]
    }
}
```

But no histogram appears in the web ui. There is nothing in the javascript console or in the terminal indicating an error.

abidlabs:
Histogram rendering was added in #309 to the old Gradio dashboard (`trackio/ui/main.py`), which displayed each histogram metric as a `gr.BarPlot` of the latest step. When the dashboard was rewritten as the Svelte SPA, that rendering path was never ported:

- The Metrics page fetches logs with `scalar_only=True`, and `SQLiteStorage._metric_rows_to_log_dicts` dropped every non-scalar value, so histogram dicts never reached the frontend.
- No frontend component handled `_type == "trackio.histogram"` (the Media page only handles image/video/audio/table).
- A project containing only histograms would not even enable the Metrics tab (`get_tab_availability_flags` and the static-mode equivalent only looked for scalar values).

So the data was written correctly, with no errors anywhere, but was silently filtered out of every read path the UI uses.

Now it's fixdd:

image

Closes: https://github.com/gradio-app/trackio/issues/637

### DIFF
--- a/.changeset/tidy-hills-cheer.md
+++ b/.changeset/tidy-hills-cheer.md
@@ -1,0 +1,5 @@
+---
+"trackio": patch
+---
+
+fix: render logged histograms in the dashboard's Metrics tab
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install ruff
+        pip install ruff==0.9.3
         
     - name: Check formatting with ruff
       run: |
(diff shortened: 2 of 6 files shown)


Saba9 (approved):
Looks great!

abidlabs:
Thanks @Saba9 for the quick review!

Issue #584, opened 2026-06-12
feature request: native (plt / plotly) Figure logging


bileamScheuvens:
W&B supports logging of Figures.

I'd prefer not lose the capability when switching.
Is that on the roadmap?
Or alternatively: what about raw html?

abidlabs:
Yes we'll add this, let me take a look at a reasonable implementation

greysonlalonde:
> Yes we'll add this, let me take a look at a reasonable implementation

Hey @abidlabs, mind if I take a crack at this?

abidlabs:
Sure thing @greysonlalonde I think the most flexible approach would be to let users log arbitrary html, using the same syntax as this: https://docs.wandb.ai/models/ref/python/data-types/html, this should allow us to handle all sorts of interactive plots.

greysonlalonde:
> Sure thing @greysonlalonde I think the most flexible approach would be to let users log arbitrary html, using the same syntax as this: https://docs.wandb.ai/models/ref/python/data-types/html, this should allow us to handle all sorts of interactive plots.

Agreed! Sounds good to me. I'll start a branch + draft.

abidlabs:
Thanks @greysonlalonde did you end up taking a stab at this? If not we might end up tackling this early next week

greysonlalonde:
> Thanks [@greysonlalonde](https://github.com/greysonlalonde) did you end up taking a stab at this? If not we might end up tackling this early next week

yup! i'll put the pr up tomorrow, afk the past couple days

abidlabs:
Closed via https://github.com/gradio-app/trackio/pull/652 thanks again @greysonlalonde

PR #652, opened 2026-07-27
feat: log arbitrary html and plotly/matplotlib figures

greysonlalonde:
adds support for #584 

Introduces `trackio.Html` for logging arbitrary html. `plotly`, `matplotlib` figures/animations passed to `trackio.log` are converted to html.

### DIFF
--- a/.changeset/html-figure-logging.md
+++ b/.changeset/html-figure-logging.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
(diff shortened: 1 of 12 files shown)


greysonlalonde:
@abidlabs ready for some eyes here 🙂

abidlabs (approved):
LGTM thanks @greysonlalonde for the nice contribution!

PR #659, opened 2026-08-04
Fix unstable metric sampling in charts that happens in `trl>=1.7.0`

### LINKED ISSUE #653: Flickering and instable charts during training when using TRL profiling.
**trackio 0.33.0**
**trl 1.9.0**

It seems like only TRL versions >= 1.7.0 are causing issues.

## Symptoms

While a run is training, line charts show gaps where points should be:

- A metric logged every `logging_steps` renders at a coarser, uneven cadence — e.g. logged every 10 steps but drawn every 20, 30, or 40, with the spacing varying along the same line.
- **The gaps move.** They jump to different x positions on every dashboard refresh (the Metrics page polls once per second), so the chart appears to flicker. Points that were there a second ago are gone; points that were missing reappear elsewhere.
- When training stops, the gaps freeze into one arbitrary pattern — sometimes a single contiguous block of missing steps, sometimes scattered dropouts.
- The data is intact on disk. Only the read path drops it, so nothing in the training logs indicates a problem.

## When it happens

Any run mixing logging cadences, once it exceeds 3000 rows in the `metrics` table. It doesn't require an unusual amount of logging — just many `log()` calls per step alongside a metric written less often.

TRL's `GRPOTrainer` reaches this quickly. `ProfilingContext._log_metrics` calls `trackio.log()` once per profiled block per step, each with a single metric, so every block becomes its own row:

```
~9 rows/step   profiling/Time taken: GRPOTrainer.*
+1 row per logging_steps   the train/* dict containing train/loss
```

`train/loss` therefore lives in roughly 1 of every ~97 rows and is maximally exposed. A 394-step run reached 3790 rows, and `train/loss` lost 7 of its 39 points, including a contiguous block of 7 consecutive steps. This surfaced after upgrading TRL 1.2.0 → 1.9.0, which added the per-block profiling logs; the same training code was fine before.

## Cause

`SQLiteStorage._subsample_metric_rows` caps each run at `max_points=3000` by decimating **whole rows uniformly by index**, without regard to which metrics each row carries. A metric occupying 1% of rows is thinned as hard as one present in every row, even though it is nowhere near the budget itself.

`trackio/sqlite_storage.py:2226`

```python
@staticmethod
def _subsample_metric_rows(rows: list[Any], max_points: int | None) -> list[Any]:
    if max_points is None or max_points < 1:
        return rows
    if len(rows) <= max_points:
        return rows
    step = len(rows) / max_points
    indices = {int(i * step) for i in range(max_points)}
    indices.add(len(rows) - 1)
    return [rows[i] for i in sorted(indices)]
```

Two defects:

1. **Not metric-aware** — sparse metrics are decimated at the same rate as dense ones.
2. **Unstable stride** — `len(rows)/max_points` is continuous and recomputed per request, so a single new row re-rounds every index and reshuffles the whole kept set. This is what makes the gaps move between refreshes.

Called from `_fetch_metric_logs_with_cursor` and `_fetch_system_logs_with_cursor`; `server.get_logs` / `get_logs_batch` pass `max_points=3000`.

## Reproduction

No GPU, no trainer, no TRL, ~1 minute:

```python
import trackio
from trackio.sqlite_storage import SQLiteStorage

trackio.init(project="repro", name="run-1")
for step in range(800):
    for i in range(9):
        trackio.log({f"dense/metric_{i}": 0.0}, step=step)
    if step % 10 == 0:
        trackio.log({"sparse/value": float(step)}, step=step)
trackio.finish()

logs = SQLiteStorage.get_logs("repro", "run-1", max_points=3000)
got = [l["step"] for l in logs if "sparse/value" in l]
print(len(got), got[:6])
```

**Expected:** `80 [0, 10, 20, 30, 40, 50]`
**Actual:** `40 [0, 20, 40, 60, 80, 100]`

Half of `sparse/value` is gone, and the chart shows a 20-step cadence for a metric logged every 10. `trackio show --project repro` renders the same thinned series. Reproduced on two independent machines.

### The moving gaps

Same layout, growing one training step at a time, tracking which `sparse/value` points survive:

```
steps=800  N=7280  kept= 40   -
steps=801  N=7290  kept= 35   39 points changed vs previous refresh
steps=802  N=7299  kept= 33   40 points changed vs previous refresh
steps=803  N=7308  kept= 33   42 points changed vs previous refresh
steps=804  N=7317  kept= 31   38 points changed vs previous refresh
```

One extra training step reshuffles nearly the entire kept set.

## Suggested fix

- Group rows by metric-key signature and allocate the budget across groups, so sparse groups keep all their points.
- Use a power-of-two stride anchored at index 0. The kept set then becomes monotone: points thin out only when the stride doubles, and never reappear at a different position between polls.

The client-side downsampler in `frontend/src/lib/dataProcessing.js` (`downsampleImpl`, 100 bins, min/max per bin) has the same instability, since bin edges depend on the current `dataXMax`. It only engages at ≥500 points per series.

## Workaround

For TRL users, suppress the per-block profiling logs before constructing the trainer. There is no config flag; profiling is gated only on `report_to`:

```python
from trl.extras import profiling
profiling.ProfilingContext._log_metrics = lambda self, duration: None
```

abidlabs:
Fixes #653.

- allocate the server-side metric row budget across metric-key signatures so sparse metrics are not discarded by unrelated dense logs
- use an anchored power-of-two stride for oversized groups, preserving stable selections as runs grow
- exclude non-scalar rows before applying the scalar chart budget
- make timestamp ordering deterministic and keep responses within the requested cap
- use power-of-two frontend bin widths so ordinary appends do not re-bin an entire series
- add backend and frontend regressions for sparse-metric preservation and append stability
- include a live browser example that crosses the 3,000-row threshold and prints an integrity check

Root cause: Trackio sampled heterogeneous SQLite rows uniformly before inspecting which metric keys each row contained. TRL profiling generates many single-metric rows, so sparse training metrics consumed only a small fraction of the row budget and were dropped even though their complete series was small. The floating sampling stride and frontend bins both depended on the current row/domain maximum, causing selected points to reshuffle on every poll.

Can verify by running: `examples/live-metric-sampling.py`

Before:

https://github.com/user-attachments/assets/3c985b9f-e9ad-457a-ac12-adc72d20f9ce

After:

https://github.com/user-attachments/assets/bd124c94-3b5c-4427-9f1e-3a6936b808fa

### DIFF
--- a/.changeset/proud-dragons-battle.md
+++ b/.changeset/proud-dragons-battle.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Fix unstable metric sampling in charts
--- a/.changeset/proud-dragons-battle.md
+++ b/.changeset/proud-dragons-battle.md
@@ -1,5 +1,5 @@
 ---
-"trackio": minor
+"trackio": patch
 ---
 
 feat:Fix unstable metric sampling in charts that happens in `trl>=1.7.0`
--- a/.changeset/proud-dragons-battle.md
+++ b/.changeset/proud-dragons-battle.md
@@ -2,4 +2,4 @@
 "trackio": minor
 ---
 
-feat:Fix unstable metric sampling in charts
+feat:Fix unstable metric sampling in charts that happens in `trl>=1.7.0`
--- a/examples/live-metric-sampling.py
+++ b/examples/live-metric-sampling.py
@@ -34,7 +34,9 @@ STEPS = int(os.environ.get("STEPS", "800"))
 STEP_DELAY = float(os.environ.get("STEP_DELAY", "0.05"))
 DENSE_METRICS = 9
 POLL_INTERVAL = float(os.environ.get("POLL_INTERVAL", "1.0"))
-PROJECT = os.environ.get("PROJECT", f"repro-653-live-{random.randint(100000, 999999)}")
+PROJECT = os.environ.get(
+    "PROJECT", f"repro-653-live-{random.randint(100000, 999999)}"
+)
 RUN_NAME = "run-1"
 OPEN_BROWSER = os.environ.get("OPEN_BROWSER", "1") != "0"
 KEEP_OPEN = os.environ.get("KEEP_OPEN", "1") != "0"
@@ -45,7 +47,11 @@ def log_step(step: int) -> None:
         trackio.log({f"dense/metric_{metric_index}": 0.0}, step=step)
 
     if step % 10 == 0:
-        loss = 2.0 + (0.75 if (step // 10) % 2 else -0.75) + 0.15 * math.sin(step / 20)
+        loss = (
+            2.0
+            + (0.75 if (step // 10) % 2 else -0.75)
+            + 0.15 * math.sin(step / 20)
+        )
         trackio.log(
             {"sparse/value": float(step), "train/loss": round(loss, 5)}, step=step
         )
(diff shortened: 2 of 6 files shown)


Saba9 (commented):
LGTM!

abidlabs:
Thanks @Saba9!

Issue #657, opened 2026-08-03
feature: support hyperparameter sweeps


Saba9:
Add ability to define, launch, and monitor hyperparameter sweeps with trackio. View results in the UI along with parallel coordinates chart.

akkiyolo:
Hi! I'd like to take a first pass at this.

Given the scope of full sweep support (launch + monitor + parallel coordinates chart) is pretty large, I'm planning to start with the backend piece as a first PR:

trackio.sweep() : define a search space (grid or random) and run it locally by calling a training function once per config
Reuses the existing config and group fields on runs, so sweep runs show up grouped in the dashboard already no schema changes
API shape roughly mirrors wandb's sweep config so existing configs mostly drop in

Not in this first PR (happy to follow up or if someone else wants to pick these up): the parallel-coordinates chart in the dashboard, a CLI launcher for external scripts, and Bayesian search.

abidlabs:
Going to close this as I believe it's not needed in Trackio, see discussion on https://github.com/gradio-app/trackio/pull/669

PR #664, opened 2026-08-11
feat: add trackio.sweep() for local grid/random hyperparameter search

akkiyolo:
Partially addresses #657.

Adds the backend piece of hyperparameter sweep support: `trackio.sweep()`
runs a local grid or random search over a defined parameter space by
calling a user-supplied training function once per configuration.

- `trackio.sweep(sweep_config, function, count=None, sweep_id=None)` ,API
  shape roughly mirrors wandb's sweep config (`method`, `parameters` with
  `values` or `min`/`max`) so existing configs mostly drop in
- Any `trackio.init()` called inside `function` automatically merges the
  sweep-selected hyperparameters into `config` and groups the run under a
  shared sweep id (unless `group` is explicitly passed), reusing the
  existing `config`/`group` fields ,no schema changes
- Added `examples/hyperparameter-sweep.py` and `tests/unit/test_sweep.py`
  (16 tests covering config generation edge cases and `init()` integration)

**Not included in this PR** (see `plans/issue-657-hyperparameter-sweeps-plan.md`
for a scoped breakdown and follow-up ideas):
- Parallel-coordinates chart in the dashboard
- CLI-driven sweeps over an external script
- Bayesian/other advanced search strategies

Ran `pytest tests/unit/test_sweep.py` (16 passed) and the full unit suite
locally with no new failures, plus `ruff check --fix --select I && ruff format`.

Commented on the issue beforehand to flag this scope:  https://github.com/gradio-app/trackio/issues/657#issuecomment-5248744380

### DIFF
--- a/examples/hyperparameter-sweep.py
+++ b/examples/hyperparameter-sweep.py
@@ -1,0 +1,46 @@
+import random
+import time
+
+import trackio as wandb
+
+
+def main():
+    project_id = random.randint(10000, 99999)
+    project_name = f"sweep-demo-{project_id}"
+    steps = 10
+
+    def train():
+        run = wandb.init(project=project_name)
+        lr = run.config["learning_rate"]
+        batch_size = run.config["batch_size"]
+
+        performance_factor = (1.0 / (lr * 100)) * (32 / batch_size)
+        base_loss = random.uniform(1.2, 1.8) / performance_factor
+        min_loss = random.uniform(0.05, 0.12) / performance_factor
+
+        for step in range(steps):
+            progress = step / (steps - 1)
+            loss = base_loss * (1.0 - 0.85 * progress) + random.uniform(-0.03, 0.03)
+            loss = max(min_loss, loss)
+            wandb.log({"loss": round(loss, 4)}, step=step)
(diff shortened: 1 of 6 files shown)


Saba9:
Hi @akkiyolo, thanks a lot for taking a crack at this, and for scoping it out in the issue beforehand!

Unfortunate timing on our end: I've been working on sweeps internally for a bit and I just pushed up a draft in #669. It covers a superset of what's here (persistent sweep/trial state in SQLite, the two-phase sweep() / agent() API matching wandb's, bayes search, remote agents, and a dashboard tab), so I don't think we can merge both — the main sticking point is the API shape. A one-shot blocking sweep(config, function) is convenient, but since trackio aims to be a drop-in wandb replacement, we need sweep() to return an id and agent() to do the running, and that's hard to walk back once released.

If you'd like, I'd genuinely appreciate a review over there — you've clearly thought about the config/parameter validation side, and a second pair of eyes on that would be great. Thanks again for the contribution.

abidlabs:
Yes let's close in favor of https://github.com/gradio-app/trackio/pull/669

PR #667, opened 2026-08-12
Add artifact registry CLI

abidlabs:
Adds a focused `trackio registry` command group on top of the registry storage and publishing APIs introduced in #645.

- create registries and typed collections
- list collections and inspect linked versions
- link local artifact versions, promote aliases, and unlink collection versions
- inspect the registry audit log
- support local and bucket-backed registries via `--bucket-id`
- provide human-readable output and `--json` for automation
- document the workflow in the registry guide and CLI reference

The CLI link path deliberately resolves local Trackio artifacts only. Artifacts that exist only on a remote Space still use the Python API, which has the context needed to record remote storage coordinates. Registry resolution/download and UI work remain separate follow-ups to #607.

cc @adrien-grl for visibility


abidlabs:
Will merge this in about ~1 hour for a minor release

PR #669, opened 2026-08-13
feat: hyperparameter sweeps (grid / random / bayes) with wandb-compatible API

### LINKED ISSUE #657: feature: support hyperparameter sweeps

Saba9:
## Short description

This PR adds hyperparameter sweep support with a wandb-compatible API. Existing wandb sweep configs work unchanged.

- **Python API**: `trackio.sweep(config)` / `trackio.agent(sweep_id, function=..., count=...)`; `trackio.init()` inside the function auto-attaches the run and merges trial params into `run.config`.
- **Search methods**: grid (atomic exactly-once cell assignment across parallel agents), random (all ~18 wandb distributions, nested parameters), and bayes (numpy-only Gaussian process + Expected Improvement — no new dependencies).
- **Command-based sweeps**: configs with `program`/`command` run trials as subprocesses with the full wandb macro set (`${env}`, `${interpreter}`, `${program}`, `${args}`, `${args_no_boolean_flags}`, `${args_no_hyphens}`, `${args_json}`, `${envvar:NAME}`), driven by a new top-level `trackio agent` CLI.
- **Lifecycle**: pause / resume / stop / cancel, `run_cap`, and `metric.target` global stop; sweeps record why they finished (`exhausted` / `run_cap` / `target`).
- **CLI**: `trackio sweep new` (JSON/TOML natively, YAML via optional pyyaml) / `list` / `status` / lifecycle commands, plus `trackio agent`.
- **Dashboard**: a Sweeps tab with trial tables, a parallel coordinates plot, best-run highlighting, and lifecycle controls; sweep runs group in the sidebar via the `_Sweep` config key. Works in static (`trackio freeze`) mode.
- **Storage**: two new SQLite tables synced through the existing parquet export/import, coordinated by the same cross-process lock as the rest of trackio — agents on different machines coordinate through a Space or self-hosted server with the existing write-token auth (no new auth).

PR does not include hyperband early termination and `run.should_stop` / killing in-flight trials on cancel. The docs in this PR state those limitations explicitly.

## Also includes

Small fixes to recently-merged registry/artifact code, prompted by review of this PR: a Spaces deadlock fix in `get_run_artifacts` (with regression test), `Artifact.unlink()` no longer applies the `TRACKIO_REGISTRY_BUCKET_ID` env fallback, server error messages now surface on 4xx responses, and fewer bucket round-trips in `BucketRegistryStorage`.

## Sweeps page
Screenshot 2026-08-18 at 10 54 25 PM

## Sweep expanded
Screenshot 2026-08-18 at 10 55 15 PM

## AI Disclosure

-----

- [x] I used AI to... implement most of this PR (Claude Code wrote the initial plan, implementation, and tests; reviewed and directed by me).
- [ ] I did not use AI

----

## Type of Change

- [ ] Bug fix
- [ ] Documentation update
- [ ] Test improvements

## Related Issues

Closes: #657

## Testing and linting

- `python -m pytest tests/unit` — 564 passed (includes ~130 new sweep tests: pure suggester/schema logic, storage lifecycle, a 4-thread grid-assignment race test, subprocess command-mode e2e, seeded bayes-beats-random check, parquet round-trip).
- `ruff check --fix --select I && ruff format` — clean.
- Frontend: `npm run build` + `npx vitest run` (147 passed).


abidlabs:
This is super cool @Saba9! I'm testing this out. Two high-level questions/suggestions:

* Do you think Sweeps should have its own dashboard (i.e. `trackio show sweeps`) like we have for the registry? I'm conflicted -- on one hand, the navbar is getting crowded but on the other hand, the sweeps ui is linked to runs, etc. Maybe we keep this as is, but consolidate some tabs (e.g. metrics and system metrics).
* Do you think we should support launching Sweeps on HF Jobs?

trackio/sqlite_storage.py
```diff
@@ -1968,6 +2096,572 @@ def get_alert_count(project: str) -> int:
...
+                    run_cap = sweep["config"].get("run_cap")
+                    reason = (
+                        "run_cap"
+                        if run_cap is not None and len(trials) >= run_cap
+                        else "exhausted"
+                    )
+                    if reason == "exhausted" and any(
+                        trial["state"] in ("assigned", "running") for trial in trials
+                    ):
+                        return {"command": "wait"}
```
abidlabs, line 2417: [P1] Reclaim trials whose agent disappeared

Once a trial has been inserted as `assigned`/`running`, nothing expires or reclaims it. If the agent process or host dies (or both terminal reports fail), a grid sweep reaches this branch with no unseen cells and returns `wait` forever; a capped sweep can instead consume the lost slot and finish short. `agent_id` and `updated_at` are recorded but never used for recovery. Please add a lease/heartbeat with expiry and fencing, or another explicit recovery path, before treating every in-flight row as permanently authoritative.

trackio/sqlite_storage.py
```diff
@@ -1159,7 +1250,10 @@ def export_to_parquet():
                     SQLiteStorage._normalize_trace_rows_for_parquet(trace_rows),
                 )
             db_mtime_ns = _sqlite_db_invalidation_mtime_ns(db_path)
-            for table in SQLiteStorage._ARTIFACT_PARQUET_TABLES:
+            for table in (
+                *SQLiteStorage._ARTIFACT_PARQUET_TABLES,
+                *SQLiteStorage._SWEEP_PARQUET_TABLES,
```
abidlabs, line 1255: [P1] Avoid overwriting grandfathered project exports

This PR explicitly preserves an existing project named, for example, `legacy_sweeps`, but its primary metrics export and the new `sweeps` sidecar for project `legacy` both resolve to `legacy_sweeps.parquet`. If both projects exist and `legacy` has sweep rows, export order/mtime makes one file overwrite or suppress the other, so import cannot recover both datasets. Please use a collision-free sidecar namespace (or migrate/reject the conflicting legacy layout before exporting sweep tables).

abidlabs (changes requested):
Adversarial review found two additional high-priority correctness/data-integrity issues. Details are inline. (The existing Copilot thread already covers the expanded-environment-secret logging issue.)

Saba9:
Thanks for the feedback @abidlabs! I'll resolve the issues you outlined and get back to you on those questions after some thought.

trackio/run.py
```diff
@@ -1711,5 +1753,67 @@ def finish(self):
                         f"They have been saved locally and will be sent automatically next time you call: "
                         f"{retry}"
                     )
+
+            self._report_sweep_trial(
```
znation, line 1757: **`_report_sweep_trial` should be in a `finally`.**

This is the last statement inside `finish()`'s broad `try` / `except Exception`, so any failure in the flush, pending-data, or bucket-drain code above it skips the trial report entirely.

Concretely: if `_drain_pending_to_bucket()` raises (e.g. a bucket credential error), the `except Exception` swallows it and `_sweep_metric_last` is never written. The parent agent's fallback then fires `report_trial(..., "finished", metric_value=None, run_id=None)` — command mode never populates those (`sweep_agent.py:353-356`) — and `metric_value = COALESCE(NULL, metric_value)` leaves the column NULL. The trial is recorded finished with no metric, so it's excluded from `_sweep_trial_summary`'s best-run query and from `target_met`: it can never win `best_run_id`, and `metric.target` can never fire on it.

`run_id` survives via `mark_trial_running`, so this is metric loss specifically. Moving the call into a `finally` on the same `try` fixes it.

trackio/asgi_app.py
```diff
@@ -321,7 +321,7 @@ async def run_api_request(request: Request, api_name: str) -> Response:
         if inspect.isawaitable(result):
             result = await result
         return JSONResponse({"data": _json_safe(result)})
-    except TrackioAPIError as e:
+    except (TrackioAPIError, ValueError) as e:
```
znation, line 324: **Broadening this to `ValueError` reclassifies every endpoint's internal errors, not just the sweep ones.**

Every `ValueError` raised anywhere in any registered endpoint is now a permanent 400.

Downstream that matters: `remote_client.py:161-167` turns a 4xx-with-error-body into `RuntimeError(error)`, and `is_transient_remote_error` (`remote_client.py:89`) returns `False` for `RuntimeError`. So the artifact-log retry loop at `run.py:1265` stops retrying and permanently drops data it previously resent — for any internal `ValueError` (a malformed-stored-data `int()`/`float()` conversion, a race) that used to surface as a retryable 500.

Suggest wrapping the sweep endpoints' errors in `TrackioAPIError` at the call site instead. `_get_registry_details` in this same PR already does exactly that, and it keeps the blast radius inside the feature.

trackio/server.py
```diff
@@ -1344,9 +1363,134 @@ def get_tab_availability(project: str) -> dict[str, bool]:
         "reports": flags["reports"] or flags["alerts"],
         "files": _project_has_files(project),
         "artifacts": flags["artifacts"],
+        "sweeps": flags["sweeps"],
     }
 
 
+def _sweep_create(
```
znation, line 1370: **The nine new sweep endpoints skip `_validate_project_name`**, which every other project-scoped endpoint in this file applies (lines 666, 679, 706, 774-825).

The motivating case isn't 500-vs-400. `POST /api/sweep_get` with `{"project": {"alpha": "beta"}}` reaches `canonical_project_name`, which iterates the dict's *keys*, canonicalizes to `"alpha"`, and silently operates on an unrelated project's sweeps. `{"project": 123}` raises `TypeError: 'int' object is not iterable` and returns a 500 carrying an internal traceback.

`_sweep_create` is partly covered because `SQLiteStorage.create_sweep` calls `validate_project_name` itself. Uncovered: `sweep_get`, `sweep_list`, `sweep_set_state`, `sweep_suggest_trial`, `sweep_mark_trial_running`, `sweep_report_trial`, `sweep_get_trials`, `sweep_run_memberships`.

trackio/sqlite_storage.py
```diff
@@ -1968,6 +2096,572 @@ def get_alert_count(project: str) -> int:
...
+                    return {"command": "exit", "reason": "target"}
+                params = sweeps_module.next_trial(sweep["config"], trials)
+                if params is None:
+                    run_cap = sweep["config"].get("run_cap")
+                    reason = (
+                        "run_cap"
+                        if run_cap is not None and len(trials) >= run_cap
+                        else "exhausted"
+                    )
+                    if reason == "exhausted" and any(
```
znation, line 2414: **An abandoned trial wedges the sweep and every agent permanently** — there's no stale-trial timeout or reclamation anywhere, so this `wait` can never resolve.

Four-cell grid sweep: agent A is assigned the last cell and is SIGKILLed before reporting. Trial 4 stays `running` forever. Every later `suggest_trial` finds `next_trial()` returns `None` (all four param hashes seen), computes `reason = "exhausted"`, sees a non-terminal trial on this line, and returns `{"command": "wait"}`. In `sweep_agent.py:341` `wait` does `time.sleep(5); continue` without incrementing `trials_run`, so `trackio.agent(sweep_id, count=10)` spins forever and never returns, and the sweep reports `running` indefinitely. The only escape is a manual `trackio sweep stop`.

Minimum real fix is a lease: reclaim `assigned`/`running` trials whose `updated_at` is older than some threshold. If that's too much for this PR, just bounding consecutive `wait`s in the agent turns a permanent wedge into a clean exit, which would be enough to unblock.

trackio/sweep_agent.py
```diff
@@ -0,0 +1,406 @@
...
+            "`trackio agent` CLI only supports command-mode sweeps."
+        )
+
+    agent_id = uuid.uuid4().hex[:8]
+    print(f"* Starting sweep agent {agent_id} for sweep {sweep_id} ({project})")
+
+    trials_run = 0
+    consecutive_failures = 0
+    while count is None or trials_run < count:
+        command = client.suggest_trial(sweep_id, agent_id=agent_id)
```
znation, line 340: **The poll loop has no transient-error resilience.** This call and the `wait` sleep sit outside the per-trial `try` that starts at line 352, so a single transport blip aborts the whole agent — while actual trial failures get three retries (`MAX_INITIAL_FAILURES`).

40-trial bayes sweep against a Space: one 502 between trials 12 and 13 (or a timeout during a sleeping-Space wake-up) raises out of this line and propagates straight out of `agent()`. The agent dies 12/40 in, even though the next poll would have succeeded. Two consecutive crashed *trials* would have been tolerated; one blip in the poll is not.

`is_transient_remote_error` already exists in `remote_client` for exactly this.

trackio/sweeps.py
```diff
@@ -0,0 +1,680 @@
...
+}
+
+
+def next_trial(
+    config: dict, trials: list[dict], rng: np.random.Generator | None = None
+) -> dict | None:
+    """Returns the next parameter set for a sweep, or None if the sweep is
+    exhausted (grid) or has hit its run_cap."""
+    run_cap = config.get("run_cap")
+    if run_cap is not None and len(trials) >= run_cap:
```
znation, line 677: **`run_cap` counts failed and pruned trials**, so crashes consume the budget.

`run_cap: 10` with 3 OOM crashes: `suggest_trial` loads all 10 rows regardless of state, this returns `None` at `len(trials) == 10`, and the sweep finishes with `finish_reason='run_cap'` having produced 7 usable results — the user asked for 10. `Sweeps.svelte` renders `{finishedTrials} / {totalTrials}` where `sweepTotalTrials` returns `run_cap`, so the row is permanently stuck at "7 / 10 finished" with a progress bar that never fills.

wandb's `run_cap` counts runs, not aborted attempts. Either count only non-failed trials here, or document the divergence.

trackio/sweeps.py
```diff
@@ -0,0 +1,680 @@
...
+        self.paths = sorted(flat_specs)
+        self.value_lists = [grid_values(flat_specs[path], path) for path in self.paths]
+
+    def suggest(
+        self, trials: list[dict], rng: np.random.Generator | None = None
+    ) -> dict | None:
+        seen = {
+            trial.get("param_hash") or param_hash(trial["params"]) for trial in trials
+        }
+        combinations = itertools.product(*self.value_lists)
```
znation, line 628: **The full grid is re-enumerated and re-hashed on every suggestion.** Each `suggest` walks the cartesian product from the start doing `unflatten_params` + `json.dumps` + SHA-256 per combination, making a grid sweep O(N^2) in hash work — all inside the cross-process sweep lock and the open SQLite connection, so every other agent and every dashboard read for the project serializes behind it.

At the `MAX_GRID_CARDINALITY` cap (100k cells) that's roughly 5x10^9 hashes over the sweep; even a 5,000-cell grid pays ~12.5M. And with `randomize_order: true`, line 631 builds a 100k-tuple list and shuffles it on *every* call.

The per-call materialize-and-shuffle is the cheap win. Persisting a resume cursor (or skipping the already-consumed prefix) makes the scan linear.

trackio/sqlite_storage.py
```diff
@@ -431,6 +458,15 @@ def _get_process_lock(project: str) -> ProcessLock:
         lockfile_path = TRACKIO_DIR / f"{canonical_project_name(project)}.lock"
         return ProcessLock(lockfile_path)
 
+    @staticmethod
+    def _get_sweep_lock(project: str) -> ProcessLock:
```
znation, line 462: **This docstring claims isolation that doesn't hold on Spaces.** "a slow trial suggestion (bayes GP fit, grid scan) never stalls metric logging" isn't true there: `suggest_trial` runs `next_trial()` inside `_get_connection`, which on Spaces unconditionally takes the single non-reentrant `_get_db_access_lock(db_path)` shared with every logging write (lines 424-435).

With ~400 completed bayes trials, `_fit_gp` does four Cholesky factorizations of a 400x400 kernel, a second fantasy refit, and 1000 candidate predictions — seconds of CPU during which every concurrent `/bulk_log`, `/get_logs`, and dashboard read for that project blocks. The separate lockfile only buys isolation *off* Spaces, i.e. not in the deployment where multi-agent sweeps are actually coordinated.

Fine by me if the perf work is deferred, but the docstring shouldn't claim it. Computing the suggestion outside the connection block (read trials, close, fit, reopen to insert) would deliver what it says.

znation (commented):
Review pass over the sweeps implementation. Eight inline comments, in the order I'd tackle them.

**Hold the merge:**
- `run.py` — `_report_sweep_trial` outside a `finally`, loses the trial's metric value
- `asgi_app.py` — `ValueError` -> 400 reclassifies every endpoint, and stops a retry loop from retrying
- `server.py` — nine sweep endpoints skip `_validate_project_name`; one input shape reads another project
- `sqlite_storage.py` — an abandoned trial wedges the sweep and every agent, permanently

**Worth folding in:**
- `sweep_agent.py` — one transient 502 kills the whole agent
- `sweeps.py` — `run_cap` counts crashed trials, so the progress bar can never complete
- `sweeps.py` — grid re-hashes the whole space per suggestion, under the lock

**Docstring only:** `_get_sweep_lock` claims isolation it doesn't provide on Spaces.

Everything else I found was cleanup or docs (a dead `finish_reason` ALTER for a table this PR introduces, an unreachable `tomli` fallback since it's already a hard dep, and `docs/source/storage_schema.md` still citing the `RegistryStorage.init_registry_db()` this PR deletes) — none of it blocking.

abidlabs:
I’ve been thinking about this PR over the last few days, and I’m wondering whether we really need a dedicated sweep API. Users haven’t asked for it, and it’s fairly trivial now to have a coding agent write a loop over configurations, create a separate Trackio run for each one, and use whichever sampling strategy makes sense. This is a very nice PR and especially UI, but given Trackio's lightweight nature,I’d lean toward documenting a few sweep recipes for now and revisiting a first-class API if we see concrete user demand

abidlabs:
Closing, but can totally reopen if you feel differently @Saba9 or @znation

PR #670, opened 2026-08-18
Improve Logbook provenance capture

abidlabs:
## Motivation

Trackio Logbooks are already good at the *human* side of reproducibility: narrative, code snippets, figures, artifacts, and live dashboards can live together on a curated page. The weak point was the machine evidence underneath that page. A manual user—or an agent—still had to remember what to attach, output was truncated, artifact detection depended on known extensions, and path-reference artifacts could change between the run and publication.

The DX goal of this work is simple: after selecting a page, the user or agent should only need to run the command through `trackio logbook run -- ...`. Trackio should capture the strongest practical record it can by default, preserve it even when the command fails, and say explicitly when a capture provider could not observe something.

### What we are learning from other provenance tools

| Inspiration | Useful idea | What this PR brings to Trackio |
| --- | --- | --- |
| [Roar](https://github.com/treqs/roar) | Treat an execution as a job that consumes and produces observed, content-identified artifacts; capture Git state and expose tracer health instead of relying on manual registration. | `logbook run` observes Python file reads/writes, records input/output lineage, snapshots files into SHA-256-addressed storage, captures Git commit/branch/dirty patch, and records provider status and limitations. Trackio artifact operations inside the command enrich that observed graph with semantic artifact identity. |
| [ReproZip](https://www.reprozip.org/) | Keep a durable execution bundle containing the command, dependencies, environment, files, and provenance needed to understand or reconstruct a run. | Every invocation gets an atomic private run bundle with the exact command, complete stdout/stderr, timestamps and exit state, runtime/package inventory, privacy-safe environment metadata, Git state, captured inputs/outputs, and integrity hashes. |
| [noWorkflow](https://github.com/gems-uff/noworkflow) | Capture useful Python provenance without requiring changes to the experiment source. | A portable `sitecustomize` audit hook captures Python runtime/import and file-access evidence automatically, including scripts running in another Python environment where Trackio itself may not be installed. |

This is intentionally the first robust capture layer, not yet a full clone of those systems. It does **not** add Roar's DAG replay/diff UI, ReproZip's OS-level dependency packer, or noWorkflow's AST/value-level provenance. Native subprocess reads and network responses remain explicit limitations. The immediate priority is high-value, cross-platform capture that works with the existing Trackio command and SDK, without adding a thicket of flags.

## Summary

- create durable, private run-evidence bundles with full stdout/stderr, Git and runtime metadata, provider diagnostics, observed Python file I/O, and content-addressed artifact snapshots
- automatically link Trackio runs and artifact lineage into the selected Logbook page when code happens to use Trackio, without requiring SDK usage for automatic file capture
- verify captured artifact digests and sizes before publishing so missing or mutated files fail loudly instead of producing incomplete Logbooks
- omit known credential files and private/cache paths by policy while recording sensitive accesses without copying their contents
- include a generic standard-library demo with zero Trackio imports that exercises stdout/stderr, input capture, and arbitrary-extension output capture using only the wrapper
- document the capture guarantees and explicit limitations, with focused coverage for failure, truncation, arbitrary outputs, renames, optional SDK linkage, privacy, and integrity checks

## Try the complete flow

The example is deliberately **not a Trackio script**. It is an ordinary Python workload with no Trackio imports or calls. From the repository root:

```sh
trackio logbook open --title "Automatic provenance capture" --no-serve
trackio logbook page "Captured training run"
trackio logbook cell markdown "This ordinary script has no Trackio instrumentation; the wrapper captures it."
trackio logbook run -- python3 examples/generic-training-demo.py
trackio logbook serve
```

Open the **Captured training run** page for a screenshot. It shows the exact command, output preview, and automatically captured `model.snapshot`, `training.metrics`, and `evaluation.json` artifacts. The private bundle additionally retains separate complete stdout/stderr streams, the input fixture and script, Git/runtime metadata, provider diagnostics, and immutable SHA-256 blobs. Its Trackio SDK provider reports `not_used`, demonstrating that the wrapper alone performed the capture.

## DX and robustness

The default `trackio logbook run` path now captures substantially more provenance without new flags or source-code changes. Full logs and all observed evidence remain in the private bundle, while the page receives a concise preview and useful links. Python workloads get runtime file-I/O observation; other workloads retain the filesystem snapshot fallback. When a workload also uses the Trackio SDK, run identity, dashboards, and semantic artifact lineage are added as an optional richer layer.

The result combines two layers that are usually separate: a compact, readable Logbook for collaboration and an immutable machine-readable evidence bundle for auditing and future reproduction work.

## Validation

- `ruff check .`
- `pytest -q -p no:rerunfailures tests/unit/test_logbook.py tests/unit/test_artifact_bucket_sync.py tests/unit/test_artifact_e2e.py tests/unit/test_artifact.py` — 102 passed
- `pytest -q -p no:rerunfailures tests/unit --ignore=tests/unit/test_gpu_hardware.py` — 434 passed
- clean end-to-end execution of the generic demo: Python audit and filesystem providers complete, Trackio SDK provider `not_used`, two inputs and three outputs captured

Not run: `tests/unit/test_gpu_hardware.py`, because this environment has no GPU, PyTorch, or pynvml.


trackio/logbook_trace.py
```diff
@@ -1173,7 +1173,9 @@ def _workspace_manifest(
     for tracked in metadata.get("local_path_artifacts") or []:
         try:
             source = Path(tracked["abs_path"]).resolve()
-            relative = source.relative_to(root).as_posix()
+            relative = Path(tracked.get("path") or source.relative_to(root)).as_posix()
```
znation, line 1176: **Three separate regressions come out of this hunk**, because `tracked["abs_path"]` now points at the CAS blob instead of the real file. (Only line 1176 is in the diff, so all three are noted here.)

**1. Line 1176 — `relative` is now cwd-relative, not root-relative.**
`tracked["path"]` is produced by `_artifact_display_path`, which is `os.path.relpath(path)` against the *process cwd*. `root` here is `proj.parent`. These differ whenever `logbook run` is invoked from a subdirectory.

> From `<repo>/experiments/`, `trackio logbook run -- python train.py` produces `experiments/out/model.pt`. The display path is `out/model.pt`, so the manifest records `path: "out/model.pt"`. `resolve_workspace_file` resolves `/__trackio_workspace__/out/model.pt` to `<repo>/out/model.pt`, finds nothing, and 404s; `sync_workspace_bucket` stats the same missing path and fails publish. The previous `source.relative_to(root)` gave the correct `experiments/out/model.pt`.

**2. Line 1185 — `source.is_file()` now tests the blob, which always exists.**
This was the guard that dropped manifest entries for files that no longer exist under the root. With `source` being the blob it can never fail.

> Run captures `out/model.pt` into `.trackio/run-evidence/blobs/...`, then the user deletes `out/`. On `trackio logbook publish` the entry survives, and `sync_workspace_bucket` builds `fingerprints` with `Path(proj.parent / "out/model.pt").stat()` (logbook_trace.py:1575) → `FileNotFoundError` → the whole publish aborts.

**3. Line 1189 — `"name": source.name` is now the blob filename.**
> `model.pt` is listed in `read_workspace_tree` and the Workspace UI as `9f2a…` (64 hex chars).

Suggest keeping `source` as the original file for the `relative` / `is_file()` / `name` derivations and carrying the blob separately, or storing a root-relative path in `local_path_artifacts` alongside `original_abs_path`.

trackio/logbook_provenance.py
```diff
@@ -0,0 +1,643 @@
...
+                            "role": role,
+                            "storage": "ignored",
+                            "reason": omission,
+                        }
+                    )
+                continue
+            try:
+                if not resolved.is_file() or resolved.is_symlink():
+                    continue
+                entry = _capture_file(self.root, resolved, cwd, role)
```
znation, line 508: **Unbounded content capture.** `_capture_file` hashes the file and then copies it whole into `.trackio/run-evidence/blobs`, for every path the audit hook observed. There is no cap on individual file size, total bytes, or file count — and each file is read twice (`cas.hash_file`, then `cas.stage_blob_from_file`).

> `trackio logbook run -- python train.py` where the script opens a 40 GB parquet shard set under the project directory: at finalize every shard is hashed and copied, doubling disk usage and reading ~80 GB, with no warning and no way to bound it. Same story for an image data loader that opens hundreds of thousands of small files.

A per-file size ceiling plus a total-bytes budget (recorded as a `limitations` entry when exceeded, in the spirit of the rest of this module) would keep the guarantee honest without the blowup.

trackio/logbook.py
```diff
@@ -2234,65 +2354,96 @@ def run_and_log(
 ) -> int:
     if not command:
         raise LogbookError("No command provided. Use: trackio logbook run -- <command>")
+    from trackio.logbook_provenance import RunEvidence  # noqa: PLC0415
+
     slug = resolve_page(proj, page)
     code_paths = _detect_code_paths(command)
+    evidence = RunEvidence.start(proj, command, Path.cwd())
```
znation, line 2361: **`--no-artifacts` and `TRACKIO_LOGBOOK_AUTONOTE=0` no longer disable capture.** `RunEvidence.start()` here and `finalize()` below run unconditionally; `capture_artifacts and _autonote_enabled()` now only gates the filesystem snapshot (line 2364) and the page cells (line 2437).

> `TRACKIO_LOGBOOK_AUTONOTE=0 trackio logbook run --no-artifacts -- python sweep.py`, following the skill doc's "Set `TRACKIO_LOGBOOK_AUTONOTE=0` to disable (e.g. during large sweeps)", still copies every file the sweep reads or writes into `.trackio/run-evidence/blobs`. The sweep pays the full disk and I/O cost the flag exists to avoid.

Either honour the existing switches for blob staging too, or give the evidence bundle its own documented opt-out.

trackio/logbook_provenance.py
```diff
@@ -0,0 +1,643 @@
...
+                        | _os.O_RDWR
+                        | _os.O_CREAT
+                        | _os.O_TRUNC
+                        | _os.O_APPEND
+                    )
+                )
+            operation = "write" if writing else "read"
+            key = (operation, path)
+            if key not in _seen:
+                _seen.add(key)
```
znation, line 322: **`_seen` and the JSONL event log grow without bound.** One tuple in `_seen` and one JSON line per distinct `(operation, path)` — plus one per distinct import — for the whole life of the child process. `_read_events` then materialises every line into a single list in the parent.

> A data loader that opens 2,000,000 distinct image files accumulates 2M tuples in `_seen` (hundreds of MB of child RSS) and writes a `python-*.jsonl` of comparable size; the wrapper then loads all 2M dicts into memory before `_stage_paths` even starts.

A bounded `_seen` (or a cap after which the hook emits a single `truncated` event and stops) would keep this proportional.

trackio/logbook_provenance.py
```diff
@@ -0,0 +1,643 @@
...
+
+    def child_environment(self) -> dict[str, str]:
+        hook_dir = self.run_dir / "python-hook"
+        hook_dir.mkdir(parents=True, exist_ok=True)
+        (hook_dir / "sitecustomize.py").write_text(
+            _sitecustomize_source(), encoding="utf-8"
+        )
+        env = os.environ.copy()
+        old_pythonpath = env.get("PYTHONPATH")
+        env["PYTHONPATH"] = os.pathsep.join(
```
znation, line 460: **The injected hook directory shadows any existing `sitecustomize.py`.** Python imports only the first `sitecustomize` on `sys.path`; prepending `hook_dir` means a pre-existing one is silently replaced rather than chained.

> On Debian/Ubuntu system Python (which ships `/usr/lib/python3/dist-packages/sitecustomize.py`), `trackio logbook run -- python train.py` runs the child with that module never imported. Its startup configuration silently stops applying, and only for commands launched through the wrapper — which makes it hard to attribute.

The generated module could `import sitecustomize` from the rest of the path after removing its own directory from `sys.path`, so the original still runs.

trackio/logbook.py
```diff
@@ -2789,7 +2971,10 @@ def _promote_local_deps(
                     print(f"  · skipped local file artifact (missing on disk): {rel}")
 
         except Exception as e:
-            print(f"  · could not push artifacts to bucket: {e}")
+            raise LogbookError(
```
znation, line 2974: **This converts every failure in the block into a fatal publish error, not just integrity failures.** The `try` starts at line 2938 and covers `create_bucket_if_not_exists`, `upload_project_to_bucket`, `_rewrite_in_pages`, and the path-artifact upload. Previously all of these printed a warning and publishing continued.

> A transient HF network error inside `create_bucket_if_not_exists` now aborts publish with "Could not publish Logbook artifacts with verified integrity" — misattributing a network fault to a digest mismatch. It also skips `write_metadata` at line 2979, and `publish()`'s rollback then restores `metadata_before_publish`, discarding the `local_dashboards[project] = target` mappings for Spaces that were *already created* above, so the retry re-promotes them.

Narrowing this to the `FileNotFoundError` / `ValueError` raised by `upload_logbook_path_artifacts_to_bucket` would get the intended fail-loud behaviour without the collateral.

trackio/logbook_provenance.py
```diff
@@ -0,0 +1,643 @@
...
+                        event = json.loads(line)
+                        if isinstance(event, dict):
+                            events.append(event)
+                    except ValueError as exc:
+                        warnings.append(
+                            f"Could not parse {path.name}:{line_number}: {exc}"
+                        )
+        except OSError as exc:
+            warnings.append(f"Could not read {path.name}: {exc}")
+    events.sort(
```
znation, line 237: **The sort key mixes two timestamp types.** The `sitecustomize` hook emits `"timestamp": time.time_ns()` (int); `record_sdk_event` emits `_now_iso()` (string). Under `str()` these compare lexicographically, so `"1755…"` always sorts before `"2026-…"`.

> Even when `trackio.init()` runs before any file access, its `trackio_run_started` event sorts after every `file_access` and `module_import` event. The ordering recorded in `run.json` — and the order in which `trackio["runs"]` entries are deduped and projected into dashboard cells — does not reflect what actually happened.

Normalising both producers to one representation (ns integers are the easier of the two) would fix it.

trackio/logbook.py
```diff
@@ -2234,65 +2354,96 @@ def run_and_log(
...
-            stderr=subprocess.STDOUT,
+            stderr=subprocess.PIPE,
             text=True,
             bufsize=1,
+            errors="replace",
+            env=evidence.child_environment(),
         )
     except FileNotFoundError as e:
+        evidence.fail_to_start(str(e))
         raise LogbookError(f"Command not found: {command[0]}") from e
```
znation, line 2384: **Only `FileNotFoundError` routes to `evidence.fail_to_start()`.** Any other `OSError` from `Popen` propagates before `finalize()` runs.

> `trackio logbook run -- ./train.sh` where `train.sh` exists but isn't executable: `Popen` raises `PermissionError`, which escapes `run_and_log`, and `.trackio/run-evidence/runs/<id>/run.json` stays at `state: "running"` / `evidence_status: "collecting"` with no `ended_at`, plus an orphaned `python-hook` directory. It is permanently recorded as an in-progress run.

Catching `OSError` here and re-raising `LogbookError` after `fail_to_start()` covers both.

znation (commented):
Correctness review of the diff — 12 findings across 10 inline comments. Overall summary posted separately in the thread.

trackio/logbook_provenance.py
```diff
@@ -0,0 +1,643 @@
...
+
+
+def _logical_path(path: Path, cwd: Path) -> str:
+    try:
+        return path.relative_to(cwd).as_posix()
+    except ValueError:
+        return path.as_posix()
+
+
+def _capture_file(root: Path, path: Path, cwd: Path, role: str) -> dict[str, Any]:
```
znation, line 180: **Inputs are hashed after the command has already run.** `_capture_file` is only reached from `finalize()`, so the recorded input digest is the file's post-run content — while the bundle presents these digests as verified provenance and `bucket_storage` later re-verifies against them.

> A script reads `config/data.csv`, then shells out to `sed -i` to rewrite it. The audit hook cannot observe the native subprocess write, so `data.csv` stays in `read_paths` and gets hashed in its already-rewritten state. `run.json` records an input digest for content that was never the input, and the publish-time integrity check passes on that wrong digest.

Worth at least recording this in the `limitations` list next to the existing native-subprocess caveat, since it is the same blind spot pointed the other way.

examples/generic-training-demo.py
```diff
@@ -0,0 +1,74 @@
...
+        encoding="utf-8",
+    )
+    metrics_path.write_text(
+        "\n".join(json.dumps(row, sort_keys=True) for row in metrics) + "\n",
+        encoding="utf-8",
+    )
+    report_path.write_text(
+        json.dumps(
+            {
+                "accuracy": accuracy,
```
znation, line 59: `accuracy` and `loss` are only bound inside the `for epoch in range(epochs)` loop but are read here and at line 43 and 60, so the script depends on the loop running at least once.

> Setting `"epochs": 0` in `examples/files/generic-training-demo.json` — a natural thing to try when exercising the demo — makes `snapshot_path.write_text(...)` raise `NameError: name 'accuracy' is not defined` instead of writing an empty-training output.

Initialising both before the loop is enough.

znation:
## Review summary

> [!NOTE]
> This review was performed by an AI agent (Claude Code). This summary and all 10 inline comments were written by it, not by a human reviewer. Please verify before acting on any of it.

Read the full diff for correctness issues. Design is sound — the audit hook, content-addressed blobs, and explicit `limitations`/provider-status fields are the right shape. **12 findings, in 10 inline comments.** Two worth blocking on:

**1. `_workspace_manifest` breaks Workspace and can fail `publish`.** Repointing `abs_path` at the CAS blob was right for integrity, but three derivations still read that field ([detail](https://github.com/gradio-app/trackio/pull/670#discussion_r3810468976)): `relative` is now cwd-relative rather than root-relative (1176), `is_file()` tests the blob so the drop-missing-files guard never fires (1185), and `name` is the blob filename so `model.pt` lists as `9f2a…` (1189). The first two reach `sync_workspace_bucket`'s unguarded `stat()` (1575) — a deleted output, or any run started from a subdirectory, now aborts the whole publish.

**2. Capture is unbounded, with no working opt-out.** `_stage_paths` byte-copies every observed read/write with no size or count cap, reading each file twice; `_seen` and the JSONL log grow one entry per distinct `(op, path)` forever. And `--no-artifacts` / `TRACKIO_LOGBOOK_AUTONOTE=0` no longer reach any of it — they now gate only the snapshot and the page cells, so the doc's "disable during large sweeps" no longer does what sweeps need.

**Smaller** — inline: broadened `raise` turns network errors into fatal integrity failures (`logbook.py:2974`); `PYTHONPATH` shadows an existing `sitecustomize.py` (`provenance.py:460`); event sort mixes ns ints with ISO strings (`:237`); `PermissionError` leaves the bundle stuck `"running"` (`logbook.py:2384`); inputs hashed post-run (`:180`); loop-leaked `accuracy`/`loss` in the demo.

Suggest fixing both clusters before merge; the rest are fine as follow-ups.

*Diff-reading only — I did not run the test suite, as no environment with the project's dependencies was available.*

PR #672, opened 2026-08-19
Add hierarchical trace UI and improve the Trackio CLI for inspecting traces

abidlabs:
## Short description

Adds the minimal agent-trace observability needed to inspect what happened inside a Trackio trace:

- stores a backwards-compatible `spans` payload across SQLite, Parquet/static exports, run renames, and run moves
- adds a hierarchical execution inspector with per-operation status, latency, cost, token usage, model, input/output, errors, and metadata
- derives tool operations from existing OpenAI-style messages when explicit spans are absent
- documents the span contract and aggregation semantics, with a realistic nested trace example and regression tests

https://github.com/user-attachments/assets/5b1d8316-7ae6-49b4-8903-093eca5023ee

Deployed here: https://abidlabs-pwc-chat-traces.hf.space/traces

### DIFF
--- a/.agents/skills/trackio/SKILL.md
+++ b/.agents/skills/trackio/SKILL.md
@@ -14,7 +14,7 @@ Trackio is an experiment tracking library for logging and visualizing ML trainin
 | **Logging metrics** during training | Python API | [logging_metrics.md](logging_metrics.md) |
 | **Firing alerts** for training diagnostics | Python API | [alerts.md](alerts.md) |
 | **Retrieving metrics & alerts** after/during training | CLI | [retrieving_metrics.md](retrieving_metrics.md) |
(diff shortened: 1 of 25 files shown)


abidlabs:
As part of this PR, I also improved the CLI and Skill around Traces, which means that you can ask your agent something like: "read these traces at https://abidlabs-pwc-chat-traces.hf.space/  and tell me how I can improve my agent" and you get concrete recommendations like this:


https://github.com/user-attachments/assets/5ad4e149-4470-4980-a45b-4aeac083840c



(thanks @NielsRogge for the suggestion)

abidlabs:
Will release this later today if anyone has time to review!

abidlabs:
I added a feature so that when using Trackio on buckets, the format renders nicely using the native agent traces viewer:

image

PR #673, opened 2026-08-20
feat: Add support for 3D objects for logging

suvadityamuk:
## Short description

This PR adds support for the logging of 3D data filetypes (specifically `GLB/gLTFs`, `PLYs`, `OBJs`, `STLs`, and `.splat`) through the introduction of a `trackio.Object3D` type along with associated tests. It also adds the ability to view them under the Media pane in a separate expandable modal pane. 

A deployed test space with what this integration looks like is [available here](https://suvadityamuk-trackio-object3d-test.hf.space/media?project=object3d-media-test).

## AI Disclosure

- [x] I used AI to draft most of the PR (using Codex). I did a review myself to cut down all the extra things it did and de-slopify it where possible. 
- [ ] I did not use AI

An AI-generated summary of changes:
```
- Main Changes
  - Added the public trackio.Object3D API:
    - Object3D(path_or_array, caption=None)
    - Object3D.from_file(path)
    - Object3D.from_numpy(array)
  - Supports .glb, .gltf, .obj, .stl, .ply, and .splat.
  - Converts NumPy inputs to binary PLY with deterministic category colors.
  - Deterministically limits ordinary point clouds to 300,000 rendered points while retaining original/rendered counts.
  - Classifies PLY files as meshes, point clouds, or Gaussian splats from their headers.

- Storage and logging
  - Uses the same upload path as images, video, and audio.
  - Supports direct logging and Object3D values nested inside tables.
  - Models must be self-contained. Embedded glTF resources are supported; external glTF resources, OBJ material libraries, and PLY textures are rejected.

- Viewer
  - Added a BabylonJS-based viewer with dynamically loaded dependencies.
  - Uses Babylon’s loaders for glTF/GLB, OBJ, STL, and Gaussian splats.
  - Added a focused ASCII/binary PLY parser for standard PLY meshes and point clouds.
  - Added lightweight cards that do not allocate WebGL until opened.
  - Cards are displayed in:
    - The Media page’s new “3D Objects” section
    - Individual table cells
    - Lists of objects inside table cells
  - Supports card selection, double-click, Enter/Space, and an explicit Open button.
  - Viewer features include:
    - Orbit, pan, and zoom
    - Automatic framing and camera reset
    - Grid and axes toggles
    - Fullscreen mode
    - Light/dark backgrounds
    - Loading and retryable error states
    - Automatic glTF animation playback
  - Disposes the Babylon scene, engine, assets, event listeners, and render loop when closed or replaced.

- Documentation and tests
[...]

PR #674, opened 2026-08-21
fix: make 3D object cards behave like image cards and fix Reset camera

abidlabs:
## Short description

Two small follow-ups to #673, ahead of the release.

### 1. 3D object cards now look and behave like image cards

Previously 3D objects had their own interaction model — a horizontal card with an explicit **Open** button and an orange "selected" border. This aligns them with images/video/audio:

- **Removed the Open button.** The preview tile is now the control itself: `cursor: zoom-in` on hover and a single click opens the viewer, exactly like clicking an image.
- **Removed the selected/highlight state**, which had no behaviour attached to it.
- **Vertical layout**, so the run name and step sit underneath the preview like video and audio.
- Uses the standard `.gallery` grid instead of a wider bespoke one.
- In table cells, renders a compact 120×80 tile sized to match `.table-image-frame`.

Accessibility improves as a side effect: the card was a `role="button"` `div` with a nested `<button>` inside it, and is now a single real `<button>`, so Enter/Space work natively rather than through a hand-rolled keydown handler.

Also drops the `Run: ` prefix from the shared media meta line — it now reads `every-media-type, Step: 5`. This affects images, video, audio, html, tables and the image modal, since the coloured run dot already conveys it.

### 2. "Reset camera" was a no-op after orbiting

Babylon's `zoomOn()` → `focusOn()` only restores `radius` and `_target`; it never touches `alpha`/`beta`. Since the initial framing already set radius and target, clicking **Reset camera** after orbiting the model appeared to do nothing at all — the rotation stayed exactly where you dragged it.

Reset now also restores the home orientation, clears any pan offset (`targetScreenOffset`) and cancels in-flight inertia.

## Type of Change

- [x] Bug fix
- [ ] Documentation update
- [x] Test improvements

## Testing and linting

`tests/ui/test_media_page.py` updated for the new interaction model (asserts the zoom-in cursor, that clicking the frame opens the viewer, and that no Open button exists), plus a new `test_reset_camera_restores_the_initial_view` that screenshots the canvas, drags to orbit, clicks Reset camera and asserts the view returns to the original pixels. **Verified that this test fails on `main`** and passes with the fix.

24 tests pass locally (`tests/ui/test_media_page.py` + `tests/unit/test_object3d.py`), along with the 131 frontend vitest tests.

### DIFF
--- a/.changeset/fancy-waves-double.md
+++ b/.changeset/fancy-waves-double.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:fix: make 3D object cards behave like image cards and fix Reset camera
--- a/tests/ui/test_media_page.py
+++ b/tests/ui/test_media_page.py
@@ -97,7 +97,9 @@ def test_media_page_shows_every_media_type(media_run):
 
         expect(page.locator(".gallery img").first).to_be_visible()
         expect(page.get_by_text("3D Objects (2)")).to_be_visible()
-        expect(page.locator(".object-card")).to_have_count(3)
+        expect(page.locator(".object-card")).to_have_count(2)
+        expect(page.locator(".object-frame.compact")).to_have_count(1)
+        expect(page.get_by_role("button", name="Open", exact=True)).to_have_count(0)
         expect(page.locator("canvas[aria-label='Interactive 3D scene']")).to_have_count(
             0
         )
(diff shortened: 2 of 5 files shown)


abidlabs:
I've verified this behaves correctly:

https://github.com/user-attachments/assets/bc55e18d-797a-4e42-abc9-5854c087c6d1

I'll go ahead and get this in for a release

PR #678, opened 2026-08-27
Fix dotted metric names in line plots

### LINKED ISSUE #677: TRL profiling metrics do not render in Trackio
## Summary

Profiling metrics emitted by TRL, such as `profiling/Time taken: GRPOTrainer.compute_loss`, generate an empty graph widget in Trackio's dashboard:

Image

I confirmed that the widgets do indeed have data by downloading their associated CSV file.

## Root Cause

I believe the culprit is the dot in the metric name. Trackio uses Vega-Lite to render the plots, and Vega-Lite treats dots as an accessor to a nested property. Trackio should escape dots in metric names.

## Details

- trackio version: 0.35.0
- trl version: 1.9.2

abidlabs:
Fixes #677.

## Summary

- Escape Vega field-path metacharacters in dynamic line-plot encodings.
- Preserve the original labels while using escaped field references for axes, series, and tooltips.
- Add regression coverage for dots, brackets, and backslashes.

## Verification

- `cd trackio/frontend && npm test`
- `cd trackio/frontend && npm run lint`
- `cd trackio/frontend && npm run build`
- `cd trackio/frontend && node repro_issue_677_vega.mjs`

## Minimal reproduction

```python
import math
import trackio

metric = "profiling/Time taken: GRPOTrainer.compute_loss"
trackio.init(project="issue-677-dotted-metric", name="trl-grpo-profile")
for step in range(12):
    trackio.log({metric: 0.25 + 0.08 * math.sin(step / 2)}, step=step)
trackio.finish()
trackio.show(project="issue-677-dotted-metric")
```

The reproduction files are intentionally untracked and were not committed.

### DIFF
--- a/.changeset/early-chairs-tan.md
+++ b/.changeset/early-chairs-tan.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Fix dotted metric names in line plots
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,18 +76,135 @@ Tests are split into unit tests (testing individual modules) and e2e tests (test
 - **Adding import formats**: Extend `imports.py` with new import functions
 - **CLI modifications**: Update `cli.py` and entry points in `pyproject.toml`
 
-## Issue Resolution Workflow
+## GitHub Issue Fix Workflow
 
-When given a GitHub issue to solve, follow this workflow:
+When asked to fix a GitHub issue and prepare a PR, use the following end-to-end
+workflow. The goal is to give reviewers observable before/after evidence, not only a
+patch and a passing unit test.
 
-1. **Create a new branch** named after the issue (e.g., `fix-issue-123` or descriptive name)
-2. **Implement the solution** following the existing code patterns and conventions
-3. **Run tests** to ensure nothing is broken: `pytest`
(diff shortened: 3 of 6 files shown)


abidlabs:
## Before / after Spaces


- [Before fix](https://huggingface.co/spaces/abidlabs/trackio-677-before) — latest released Trackio
- [After fix](https://huggingface.co/spaces/abidlabs/trackio-677-after) — preview wheel from this PR

| Space | Expected result |
|---|---|
| Before | `profiling/reference_metric` renders; `profiling/Time taken: GRPOTrainer.compute_loss` is empty |
| After | Both profiling metrics render |

Both Spaces use byte-identical `app.py` files and serve the same 12 values for each metric. The after Space serves the exact frontend bundle built from this branch.

qgallouedec (approved):
Not a trackio expert but this looks right to me. I checked `escapeVegaField` against vega-util's `splitAccessPath` and it round-trips `profiling/Time taken: GRPOTrainer.compute_loss` back to the literal key, and every `field:` in the spec looks covered.

One thing I wasn't sure about: quotes aren't escaped. `splitAccessPath` seems to enter quote mode whenever `i === b`, which for a field with no dots holds at every position, so something like `say "hi"` would still break. vega-lite's `replacePathInField` uses `/(\[|\]|\.|'|")/g`. Might be worth the union, though it doesn't affect us, our metric names come from Python identifiers.

Thanks for the quick turnaround.

abidlabs:
Thanks for catching that @qgallouedec! Addressed now.

PR #680, opened 2026-08-30
fix: decode UTF-8 blobs in query output instead of hex-encoding

B1aZer:
JSON columns (`config`, `metrics`, ...) are stored as orjson-encoded UTF-8 BLOBs, but `_normalize_query_value` hex-encodes every bytes value, so the documented query interface returns unreadable output in both table and `--json` modes:

```
$ trackio query project --project my-project --sql "SELECT config FROM configs LIMIT 1"
7b226c72223a302e30312c...
```

This is especially confusing for agents using the query path that `storage_schema.md` and the trackio skill document as a first-class access method.

**Change:** try UTF-8 decode first; fall back to hex only for genuinely binary data.

**Tests:**
- new `test_query_project_decodes_utf8_blobs` — a stored config round-trips through `query_project` as parseable JSON text
- `test_query_project_normalizes_bytes` now uses a fixed invalid-UTF-8 blob (`x'80FF8081'`) — the previous `randomblob(4)` could occasionally produce valid UTF-8 and flake under this change

All 32 tests in `tests/unit/test_sqlite_storage.py` pass locally. Found while validating the SQL recipes for the sweep docs PR.

### DIFF
--- a/.changeset/query-utf8-blobs.md
+++ b/.changeset/query-utf8-blobs.md
@@ -1,0 +1,5 @@
+---
+"trackio": patch
+---
+
+fix: decode UTF-8 blobs in `trackio query` output instead of hex-encoding them, so JSON columns like `config` and `metrics` are readable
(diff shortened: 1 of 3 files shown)


abidlabs (approved):
Thank you for the fix @B1aZer!

PR #681, opened 2026-08-30
docs: add hyperparameter sweep recipes

B1aZer:
Follows up on #657 and the discussion in #669, where @abidlabs suggested documenting sweep recipes rather than adding a first-class sweep API.

This adds:

- `docs/source/sweeps.md` — recipes for **grid search**, **random search**, and **Bayesian optimization via Optuna**, all built on the existing `init(group=..., config=...)` pattern; a section on viewing sweep results in the dashboard; a section on querying the best configuration over SQL (including that a run's group is stored under the `_Group` key in the config JSON); and a short section on delegating sweeps to coding agents, linking to the ML Agents guide.
- `examples/sweep-recipes.py` — runnable, dependency-free example with the grid and random recipes.
- A toctree entry under How-to guides.

Every snippet was run against a local install from `main`, including the Optuna recipe. While validating the SQL recipe I found that `SELECT config ...` comes back hex-encoded through `trackio query` — fix proposed separately in #680.

Happy to adjust scope or tone — e.g. trim the Optuna section if external-library recipes don't belong in the docs.

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -11,6 +11,8 @@
 - sections:
   - local: track
     title: Track
+  - local: sweeps
+    title: Hyperparameter Sweeps
   - local: traces
     title: Traces
   - local: artifacts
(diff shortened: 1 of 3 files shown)


abidlabs (approved):
I think this is worth adding to the docs so that users/llms can easily find it, thanks @B1aZer!

PR #684, opened 2026-09-04
Reduce dashboard canvas GPU memory and render churn

### LINKED ISSUE #683: the browser tab consumes a ton of VRAM (Brave - Windows)
It's all in the title. The browser tab eats sometimes about 2GB of VRAM, the interface reacts very slowly sometimes etc.

That's obviously the only tab that I has this behavior and it's not browser or machine-wide.

abidlabs:
Issue #683 reports browser-tab GPU memory reaching roughly 2 GB and severe UI slowdown. The dashboard previously kept a canvas-backed Vega view mounted for every metric chart, including charts far outside the viewport. Async embeds and interval refreshes could also overlap, allowing stale rendering work and redundant requests to accumulate.

Now instead, we display the plots when they are visible to the user. Here's what it looks like for reference:

https://github.com/user-attachments/assets/babf6fe4-290b-4b49-85fc-4f9921c97538

Closes: #683

### DIFF
--- a/.changeset/all-donkeys-drum.md
+++ b/.changeset/all-donkeys-drum.md
@@ -1,0 +1,5 @@
+---
+"trackio": minor
+---
+
+feat:Reduce dashboard canvas GPU memory and render churn
--- a/.changeset/all-donkeys-drum.md
+++ b/.changeset/all-donkeys-drum.md
@@ -1,5 +1,5 @@
 ---
-"trackio": minor
+"trackio": patch
 ---
 
 feat:Reduce dashboard canvas GPU memory and render churn
--- a/tests/ui/test_ui_display.py
+++ b/tests/ui/test_ui_display.py
@@ -201,3 +201,39 @@ def test_multiple_runs_display_multiple_plots(temp_dir):
     finally:
         trackio.delete_project("test_multi", force=True)
         app.close()
+
+
+def test_metric_charts_release_canvases_outside_viewport(temp_dir):
+    trackio.init(project="test_chart_visibility", name="run")
+    for step in range(2):
+        trackio.log(metrics={f"metric_{i}": i + step for i in range(20)})
+    trackio.finish()
+
+    app, _, _, full_url = trackio.show(
+        project="test_chart_visibility", block_thread=False, open_browser=False
+    )
+
+    try:
+        with sync_playwright() as p:
+            browser = p.chromium.launch()
+            page = browser.new_page(viewport={"width": 1280, "height": 720})
+            page.set_default_timeout(10000)
+            page.goto(full_url)
+
+            plots = page.locator(".metrics-page .plot")
+            expect(plots).to_have_count(20)
+            expect(plots.first.locator("canvas")).to_have_count(1)
+
+            initial_canvas_count = page.locator(".vega-embed canvas").count()
+            assert 0 < initial_canvas_count < 20
+
+            page.locator(".metrics-page").evaluate(
+                "element => { element.scrollTop = element.scrollHeight; }"
+            )
+            expect(plots.last.locator("canvas")).to_have_count(1)
+            expect(plots.first.locator("canvas")).to_have_count(0)
+
+            browser.close()
+    finally:
+        trackio.delete_project("test_chart_visibility", force=True)
+        app.close()
--- a/tests/ui/test_ui_display.py
+++ b/tests/ui/test_ui_display.py
@@ -206,7 +206,7 @@ def test_multiple_runs_display_multiple_plots(temp_dir):
 def test_metric_charts_release_canvases_outside_viewport(temp_dir):
     trackio.init(project="test_chart_visibility", name="run")
     for step in range(2):
-        trackio.log(metrics={f"metric_{i}": i + step for i in range(20)})
+        trackio.log(metrics={f"metric_{i}": i + step for i in range(48)})
     trackio.finish()
 
     app, _, _, full_url = trackio.show(
@@ -221,15 +221,33 @@ def test_metric_charts_release_canvases_outside_viewport(temp_dir):
             page.goto(full_url)
 
             plots = page.locator(".metrics-page .plot")
-            expect(plots).to_have_count(20)
+            expect(plots).to_have_count(48)
             expect(plots.first.locator("canvas")).to_have_count(1)
+            page.wait_for_function(
+                "() => document.querySelectorAll('.vega-embed canvas').length < 48"
+            )
 
             initial_canvas_count = page.locator(".vega-embed canvas").count()
-            assert 0 < initial_canvas_count < 20
+            initial_first_height = plots.first.evaluate(
+                "element => element.getBoundingClientRect().height"
+            )
+            assert 0 < initial_canvas_count < 48
 
             page.locator(".metrics-page").evaluate(
                 "element => { element.scrollTop = element.scrollHeight; }"
             )
+            expect(plots.last.locator("canvas")).to_have_count(1)
+            expect(plots.first.locator("canvas")).to_have_count(0)
+            released_first_height = plots.first.evaluate(
+                "element => element.getBoundingClientRect().height"
+            )
+            scroll_position = page.locator(".metrics-page").evaluate(
+                "element => ({ top: element.scrollTop, max: element.scrollHeight - element.clientHeight })"
+            )
+            assert abs(released_first_height - initial_first_height) < 1
+            assert abs(scroll_position["top"] - scroll_position["max"]) < 10
+
+            page.set_viewport_size({"width": 720, "height": 900})
             expect(plots.last.locator("canvas")).to_have_count(1)
             expect(plots.first.locator("canvas")).to_have_count(0)
 
--- a/tests/ui/test_ui_display.py
+++ b/tests/ui/test_ui_display.py
@@ -223,9 +223,7 @@ def test_metric_charts_release_canvases_outside_viewport(temp_dir):
             plots = page.locator(".metrics-page .plot")
             expect(plots).to_have_count(48)
             expect(plots.first.locator("canvas")).to_have_count(1)
-            page.wait_for_function(
-                "() => document.querySelectorAll('.vega-embed canvas').length < 48"
-            )
+            expect(plots.last.locator("canvas")).to_have_count(0)
 
             initial_canvas_count = page.locator(".vega-embed canvas").count()
             initial_first_height = plots.first.evaluate(
@@ -248,6 +246,9 @@ def test_metric_charts_release_canvases_outside_viewport(temp_dir):
             assert abs(scroll_position["top"] - scroll_position["max"]) < 10
 
             page.set_viewport_size({"width": 720, "height": 900})
+            page.locator(".metrics-page").evaluate(
+                "element => { element.scrollTop = element.scrollHeight; }"
+            )
             expect(plots.last.locator("canvas")).to_have_count(1)
             expect(plots.first.locator("canvas")).to_have_count(0)
 
--- a/trackio/frontend/src/components/LinePlot.svelte
+++ b/trackio/frontend/src/components/LinePlot.svelte
@@ -4,6 +4,10 @@
   import * as vega from "vega";
   import { buildColorSpecKey } from "../lib/dataProcessing.js";
   import { visibleLegendEntries } from "../lib/legend.js";
+  import {
+    createVegaViewManager,
+    observeNearViewport,
+  } from "../lib/chartLifecycle.js";
 
   let {
     data = [],
@@ -30,8 +34,9 @@
   let container = $state(null);
   let plotContainer = $state(null);
   let fullscreenHost = $state(null);
-  let view = $state(null);
   let fullscreen = $state(false);
+  let nearViewport = $state(false);
+  const viewManager = createVegaViewManager();
 
   let lastStructuralKey = null;
   let lastHasSmoothed = false;
@@ -317,6 +322,7 @@
   }
 
   function tryIncrementalUpdate() {
+    const view = viewManager.current;
     if (!view) return false;
 
     const { originalData, smoothedData, hasSmoothed } = splitData();
@@ -340,6 +346,7 @@
   }
 
   function syncViewSize() {
+    const view = viewManager.current;
     if (!view || !container) return;
     view.width(container.clientWidth);
     if (fullscreen) view.height(container.clientHeight);
@@ -350,20 +357,27 @@
     await tick();
     if (!container || !data || data.length === 0 || !y) return;
 
+    const target = container;
     const spec = buildSpec();
+    const structuralKey = getStructuralKey();
 
     try {
-      if (view) {
-        view.finalize();
-        view = null;
+      const result = await viewManager.replace(
+        () =>
+          embed(target, spec, {
+            actions: false,
+            renderer: "canvas",
+          }),
+        target,
+      );
+      if (!result || target !== container) {
+        if (result) viewManager.clear();
+        return;
       }
-      const result = await embed(container, spec, {
-        actions: false,
-        renderer: "canvas",
+      lastStructuralKey = structuralKey;
+      requestAnimationFrame(() => {
+        if (viewManager.current === result.view) syncViewSize();
       });
-      view = result.view;
-      lastStructuralKey = getStructuralKey();
-      requestAnimationFrame(syncViewSize);
 
       if (onSelect) {
         let lastSelectTime = 0;
@@ -386,10 +400,14 @@
   }
 
   async function render() {
-    if (!container || !data || data.length === 0 || !y) return;
+    if (!nearViewport || !container || !data || data.length === 0 || !y) {
+      viewManager.clear();
+      lastStructuralKey = null;
+      return;
+    }
 
     const structuralKey = getStructuralKey();
-    if (view && structuralKey === lastStructuralKey) {
+    if (viewManager.current && structuralKey === lastStructuralKey) {
       if (tryIncrementalUpdate()) return;
     }
 
@@ -425,6 +443,7 @@
   }
 
   async function downloadImage() {
+    const view = viewManager.current;
     if (!view) return;
     try {
       const url = await view.toImageURL("png", 4);
@@ -534,7 +553,15 @@
     title;
     fullscreen;
     container;
+    nearViewport;
     render();
+  });
+
+  $effect(() => {
+    if (!container) return;
+    return observeNearViewport(container, (visible) => {
+      nearViewport = visible;
+    });
   });
 
   $effect(() => {
@@ -556,7 +583,7 @@
       document.removeEventListener("webkitfullscreenchange", onFullscreenChange);
       document.removeEventListener("mozfullscreenchange", onFullscreenChange);
       document.removeEventListener("MSFullscreenChange", onFullscreenChange);
-      if (view) view.finalize();
+      viewManager.destroy();
       document.body.style.overflow = "";
     };
   });
@@ -942,6 +969,7 @@
   }
   .plot {
     width: 100%;
+    min-height: 250px;
   }
   .plot :global(.vega-embed) {
     width: 100% !important;
--- a/trackio/frontend/src/components/LinePlot.svelte
+++ b/trackio/frontend/src/components/LinePlot.svelte
@@ -981,7 +981,7 @@
   }
   .plot {
     width: 100%;
-    min-height: 250px;
+    min-height: 300px;
   }
   .plot :global(.vega-embed) {
     width: 100% !important;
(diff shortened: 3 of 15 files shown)


trackio/frontend/src/pages/SystemMetrics.svelte
```diff
@@ -265,31 +267,35 @@
...
+        console.error("Failed to refresh system metric logs:", e);
+      }
     }
   }
 
   $effect(() => {
     project;
     selectedRuns;
     appBootstrapReady;
+    refreshTask.cancel();
```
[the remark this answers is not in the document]
abidlabs: Fixed in 8afd2f1. System-metric initial and selection loads now share the abortable polling task, including batched and per-run fallback requests, with stale-project response guards.

trackio/frontend/src/pages/Metrics.svelte
```diff
@@ -265,31 +272,35 @@
...
+        console.error("Failed to refresh metric logs:", e);
+      }
     }
   }
 
   $effect(() => {
     project;
     selectedRuns;
     appBootstrapReady;
+    refreshTask.cancel();
```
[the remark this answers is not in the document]
abidlabs: Fixed in 8afd2f1. Initial and selection loads now run through the same abortable polling task, propagate the signal through batched requests, and ignore responses after the project changes.

znation (commented):
## 🤖 AI-generated code review

**This review was generated by Claude Code (Claude Opus 5) and posted by me on its behalf. It has not been hand-verified line by line.** Please treat every comment as a hypothesis to check rather than a confirmed defect. In particular, nothing here is backed by an actual failing run — `trackio/frontend/node_modules` was not installed in the review checkout, so `npm test` / `npm run lint` / `npm run build` and the new Playwright test were never executed. Everything below comes from reading the diff and the surrounding code paths.

Overall this looks like a solid change, and the approach — an `IntersectionObserver` gate plus a generation-guarded view manager, with real unit tests for the manager — seems like the right shape for #683. Submitting as a comment, not a change request.

Six items, roughly in the order I'd prioritise them:

| # | Item | Where |
|---|------|-------|
| 1 | Aborted first load can leave the dashboard stuck on the spinner (the one I'd call merge-blocking) | `Metrics.svelte`, `api.js` |
| 2 | `observeNearViewport` reads only the first `IntersectionObserver` entry | `chartLifecycle.js` |
| 3 | Manual `.changeset` file, against repo `CLAUDE.md` | `.changeset/` |
| 4 | `min-height: 250px` under-reserves the real chart height | `LinePlot.svelte` +2 |
| 5 | Two problems with the new Playwright test | `test_ui_display.py` |
| 6 | A failed `embed()` orphans the previous chart's canvas | `chartLifecycle.js` |

Details inline.

trackio/frontend/src/pages/Metrics.svelte
```diff
@@ -236,20 +247,28 @@
...
     let fetched = false;
     if (needFetch.length > 0) {
       try {
-        const batch = await fetchLogsForRuns(needFetch);
+        const batch = await fetchLogsForRuns(
+          needFetch,
+          { signal },
+          requestedProject,
+        );
+        if (signal.aborted || requestedProject !== project) return;
```
znation, line 264: ### 1. An aborted first load can leave the dashboard on the spinner forever

`createPollingTask()` defaults to a 30s hard timeout, so the initial `refreshTask.run(fetchNewRuns)` gets aborted whenever the first `get_logs_batch` takes longer than that — a large project, a cold Space, or slow static parquet reads.

When that happens `fetchNewRuns` returns here (or at the `if (e?.name === "AbortError") return;` a few lines down) and never reaches `hasLoaded = true` on line 279. The only thing that can recover is `refreshCachedRuns`, which bails immediately when `!realtimeEnabled` — and `App.svelte:453` sets `realtimeEnabled = false` for **every static-mode dashboard**. So a static dashboard whose first fetch exceeds 30s renders `<LoadingTrackio />` indefinitely, with no error and no retry.

Before this PR there was no timeout at all, and any failure still fell through to `hasLoaded = true`.

`SystemMetrics.svelte:257` / `:265` has the identical structure.

Two suggestions, in order of how much I'd want them:

- set `hasLoaded = true` before returning on the abort paths, so the worst case is the empty state rather than an infinite spinner;
- reconsider applying a 30s cap to the *initial* load at all. Cancel-on-supersede is the valuable half of `createPollingTask`; the timeout is separable, and as written it's a brand-new hard failure mode for exactly the large-log-volume projects #683 came from.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

trackio/frontend/src/lib/api.js
```diff
@@ -94,7 +95,12 @@ export async function getLogs(project, run, options = {}) {
...
 }
 
-export async function getLogsBatch(project, runs, options = {}) {
+export async function getLogsBatch(
+  project,
+  runs,
+  options = {},
+  requestOptions = {},
+) {
   if (await isStaticMode()) {
```
znation, line 104: ### 1b. The static-mode branches drop `requestOptions`, so the abort can't cancel anything

Companion to the comment on `Metrics.svelte`. `getLogsBatch` takes a `requestOptions` argument now, but this branch loops `staticApi.getLogs(project, run, options)` without the signal — same in `getSystemLogsBatch` and `getSystemLogs` further down.

Net effect on a static dashboard: when the 30s timer fires, the N sequential parquet reads keep running to completion (so none of the bandwidth or CPU the abort was meant to save is actually saved), and then the caller discards the fully-fetched result because `signal.aborted` is true. That's what turns the timeout into the stuck-spinner failure rather than a harmless no-op.

Either thread the signal into `staticApi`, or skip the timeout entirely when `isStaticMode()`.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

trackio/frontend/src/lib/chartLifecycle.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+export function observeNearViewport(element, onChange) {
+  if (typeof IntersectionObserver === "undefined") {
+    onChange(true);
+    return () => {};
+  }
+
+  const observer = new IntersectionObserver(
+    ([entry]) => onChange(entry?.isIntersecting ?? false),
```
znation: ### 2. Only the first `IntersectionObserver` entry is read

`IntersectionObserver` hands the callback an *array* of records in chronological order, and more than one record for the same target arrives whenever visibility flips twice inside a single observation cycle — fast or momentum scrolling, or a resize plus a scroll in the same frame.

Destructuring `([entry])` takes the **oldest** record, so `nearViewport` gets set to a stale value and nothing re-fires until the next genuine change. Concretely: a chart the user is looking at can sit blank at its 250px placeholder, or an off-screen chart can keep its canvas — either way the mechanism this PR is built on silently mis-fires.

```suggestion
    (entries) => onChange(entries[entries.length - 1]?.isIntersecting ?? false),
```

Worth extending the `observeNearViewport` unit test with a single batched call — `callback([{ isIntersecting: false }, { isIntersecting: true }])` — to pin the behaviour.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

.changeset/all-donkeys-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"trackio": patch
+---
+
+feat:Reduce dashboard canvas GPU memory and render churn
```
znation, line 5: ### 3. Manual changeset — repo `CLAUDE.md` says not to add one

> Don't create a .changeset file manually in the PR. Once you create a PR, a GH action will create it automatically based on the PR title

The bot will push its own changeset derived from the PR title, so this one yields a duplicate release-note entry, and the bot commit will reject the next non-rebased push to the branch.

Minor extra: this file is tagged `patch` while the summary is prefixed `feat:` (and there's no space after the colon), which is not the shape the action generates.

Suggest simply deleting the file.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

trackio/frontend/src/components/LinePlot.svelte
```diff
@@ -950,6 +981,7 @@
   }
   .plot {
     width: 100%;
+    min-height: 250px;
```
znation: ### 4. `min-height: 250px` under-reserves the height the chart will actually take

`buildSpec` uses `height: fullscreen ? "container" : 250` (line 281), but Vega's `height` is the *inner* mark height — axes, tick labels and the axis title add roughly another 40–60px, so a rendered `.plot` lands nearer 300px.

With 48 metrics only the ~18 inside the 600px `rootMargin` are mounted; the other ~30 sit at 250px each, under-reserving well over 1000px of scroll height. As you scroll, every chart that mounts grows ~50px and pushes content down under the scroll position — cumulative layout shift and a scrollbar that keeps changing length. It's the most visible cost of the change, and it partly undoes the no-layout-jump property the held-height logic exists to provide.

The new Playwright test only covers the held-height path for a chart that *had* already rendered, so it won't catch this.

Options: derive the placeholder from the same constant as the spec plus measured chrome, or capture one real rendered height and reuse it. Same line in `BarPlot.svelte:632` and `HistogramPlot.svelte:574` — the `250` is now spread across six places in three components.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

tests/ui/test_ui_display.py
```diff
@@ -201,3 +201,57 @@ def test_multiple_runs_display_multiple_plots(temp_dir):
...
+        with sync_playwright() as p:
+            browser = p.chromium.launch()
+            page = browser.new_page(viewport={"width": 1280, "height": 720})
+            page.set_default_timeout(10000)
+            page.goto(full_url)
+
+            plots = page.locator(".metrics-page .plot")
+            expect(plots).to_have_count(48)
+            expect(plots.first.locator("canvas")).to_have_count(1)
+            page.wait_for_function(
```
znation: ### 5a. This `wait_for_function` is trivially true and waits for nothing

By the time it runs, the preceding `expect(plots.first.locator("canvas")).to_have_count(1)` already guarantees one canvas, and `1 < 48` passes on the very first poll. So `initial_canvas_count` on the next line is snapshotted while charts are still mounting.

That matters because `assert 0 < initial_canvas_count < 48` would very likely *also* pass on the pre-fix build, for the same reason — the test can't reliably fail on the regression it's meant to guard.

Suggest waiting for the count to settle before snapshotting (poll until two consecutive readings are equal), or asserting on a specific plot that should be well outside the margin, e.g. `expect(plots.nth(47).locator("canvas")).to_have_count(0)`.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

tests/ui/test_ui_display.py
```diff
@@ -201,3 +201,57 @@ def test_multiple_runs_display_multiple_plots(temp_dir):
...
+            released_first_height = plots.first.evaluate(
+                "element => element.getBoundingClientRect().height"
+            )
+            scroll_position = page.locator(".metrics-page").evaluate(
+                "element => ({ top: element.scrollTop, max: element.scrollHeight - element.clientHeight })"
+            )
+            assert abs(released_first_height - initial_first_height) < 1
+            assert abs(scroll_position["top"] - scroll_position["max"]) < 10
+
+            page.set_viewport_size({"width": 720, "height": 900})
```
znation, line 248: ### 5b. This resize step will probably flake in CI

The grid is `--plot-min-width: 300px` with `--plot-max-cols: 4`, so going 1280 → 720 roughly halves the column count and roughly doubles the grid height. `scrollTop` is still pinned to the *old* `scrollHeight - clientHeight`, so it is no longer at the bottom of the new layout, and whether `plots.last` lands inside the 600px `rootMargin` comes down to Chromium's scroll-anchoring heuristics.

I'd expect `expect(plots.last.locator("canvas")).to_have_count(1)` on the next line to time out intermittently. Re-scrolling to the bottom after the resize would make it deterministic.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

trackio/frontend/src/lib/chartLifecycle.js
```diff
@@ -0,0 +1,145 @@
...
+    if (!queued) return;
+    queued.resolve(null);
+    queued = null;
+  }
+
+  async function execute(request) {
+    if (destroyed || request.generation !== generation) return null;
+
+    finalizeView(current);
+    if (currentElement && currentElement !== request.element) {
```
znation, line 46: ### 6. A failed `embed()` orphans the previous chart's canvas

On a re-render into the same container this condition is false, so the old DOM is deliberately kept — fine, `vega-embed` replaces it. But line 45 has already finalized the old view, and lines 49–50 then null both `current` and `currentElement` *before* awaiting `request.create()`.

If `create()` rejects — a bad spec from a malformed metric, or canvas/WebGL context exhaustion, which is exactly the condition #683 describes — or resolves without a `view` (the `throw` on line 55), the element is left holding the finalized view's `<canvas>` and the manager has no reference to it any more. `currentElement` is `null`, so neither `clear()` nor `destroy()` will ever `replaceChildren()` it: the dead canvas stays in the DOM and its GPU memory stays pinned until the component unmounts. `releaseElementHeight` is skipped too, so any previously held inline height stays frozen.

Narrow trigger, but it's a couple of lines in brand-new code, and the failure mode is precisely what this PR exists to prevent. Cheapest fix is probably to leave `currentElement` set until the create succeeds, or to clear `request.element` from a `catch` around the await.

---
*🤖 AI-generated review comment (Claude Code / Opus 5). Not verified by a human, and not backed by a failing test run — please sanity-check before acting.*

abidlabs:
Thank you @znation! Addressed the important findings and pushed

trackio/frontend/src/pages/Metrics.svelte
znation: Addressed in db01ea9 — `createPollingTask.run()` now takes a per-run timeout, and the initial `refreshTask.run(fetchNewRuns, null)` opts out of it entirely, so the first load can no longer be aborted by the watchdog. Every remaining `cancel()` is immediately followed by a fresh `run(fetchNewRuns, null)`, so the stranded-`hasLoaded` state is no longer reachable. Resolving.

One residual nit, not worth blocking on: the `if (signal.aborted) return;` and `if (e?.name === "AbortError") return;` paths still skip `hasLoaded = true`, so the invariant now rests entirely on "every cancel is followed by a new run". Setting `hasLoaded = true` before those returns would keep it robust if a timeout or another cancel path is ever reintroduced.

*🤖 AI-generated (Claude Code).*

trackio/frontend/src/lib/api.js
znation: Addressed in db01ea9 — the signal is now threaded through `staticApi.getLogs`/`getSystemLogs` → `getMetricsData`/`getSystemData` → `readParquet(url, {}, signal)`, with a `throwIfAborted()` fast path in front. Resolving.

Worth knowing for later: `readParquet` only populates its cache on success, so an aborted parquet download is now discarded entirely rather than completing into the cache. That's the right trade for cancel-on-supersede, but it does mean a `cancel()` mid-download restarts from scratch. Bounded in practice — static run lists are stable, so the effect settles after bootstrap.

*🤖 AI-generated (Claude Code).*

trackio/frontend/src/lib/chartLifecycle.js
znation: Fixed exactly as suggested in db01ea9, and the unit test now feeds a batched `callback([{ isIntersecting: false }, { isIntersecting: true }])`, which genuinely fails against the old `([entry])` destructure. Confirmed this actually executes in CI — `.github/workflows/test.yml` runs `npm test` in the `test (hf-jobs-cpu-upgrade)` job, which is green on this head. Resolving.

*🤖 AI-generated (Claude Code).*

.changeset/all-donkeys-drum.md
znation: **Retracting this one — it was wrong.** I checked authorship: `.changeset/all-donkeys-drum.md` was committed by `gradio-pr-bot` in 20aeb89, which is exactly the automated flow `CLAUDE.md` describes. Nothing here was added by hand, and the `feat:` prefix and spacing are the bot's own generated format from the PR title, not an authoring mistake.

My comment came from seeing the file in the diff against `main` without checking who committed it. Apologies for the noise — resolving as invalid.

*🤖 AI-generated (Claude Code).*

trackio/frontend/src/components/LinePlot.svelte
znation: Addressed in db01ea9 — bumped to 300px in all three components, which lines up with the ~250px spec height plus axis chrome. Resolving.

Caveat worth stating plainly: I haven't measured the rendered height in a browser, so 300 is a plausible value rather than a verified one — worth an eyeball at a couple of viewport widths. The constant is still duplicated across the three components, but that de-duplication was always follow-up work rather than part of this PR.

*🤖 AI-generated (Claude Code).*

tests/ui/test_ui_display.py
znation: Addressed in db01ea9 — replaced with `expect(plots.last.locator("canvas")).to_have_count(0)`, which is a real assertion and would fail on the pre-fix build where all 48 charts mount. That's a better fix than the one I suggested. Resolving.

Minor residual, not worth another round-trip: `to_have_count(0)` is satisfied on the first poll, so in principle it could still pass before the last chart would have rendered.

*🤖 AI-generated (Claude Code).*

tests/ui/test_ui_display.py
znation: Addressed in db01ea9 — re-scrolls to the bottom after the resize, so the assertion no longer depends on Chromium's scroll-anchoring heuristics. The `ui/ux interaction tests` job is green on this head. Resolving.

*🤖 AI-generated (Claude Code).*

znation (approved):
## 🤖 AI-generated review — approving

Same disclaimer as before: this was generated by **Claude Code (Claude Opus 5)** and posted by me on its behalf; it has not been hand-verified line by line.

db01ea9 addresses everything substantive from the previous round. Re-checked against the current head:

| # | Item | Status |
|---|------|--------|
| 1 | Stuck spinner on an aborted first load | ✅ Fixed — per-run timeout, initial load opts out via `run(fetchNewRuns, null)` |
| 1b | Static branches dropped the abort signal | ✅ Fixed — threaded through to `readParquet` |
| 2 | `entries[0]` in the observer callback | ✅ Fixed, plus a batched-entries regression test |
| 3 | "Manual" changeset | ❌ **My finding was wrong** — bot-authored in 20aeb89. Retracted. |
| 4 | `min-height` under-reserved the chart height | ✅ 250px → 300px in all three components |
| 5a | No-op `wait_for_function` | ✅ Replaced with a real assertion |
| 5b | Flaky resize assertion | ✅ Re-scrolls after the resize |
| 6 | Orphaned canvas on a failed `embed()` | ⏳ Still open — non-blocking, see below |

I've resolved threads 1–5 and left 6 open.

### Correction on #3

I claimed the changeset was added by hand in violation of `CLAUDE.md`. It wasn't — `gradio-pr-bot` committed it in 20aeb89, which is exactly the automated flow the doc describes. I'd seen the file in the diff against `main` without checking authorship. Retracted, and sorry for the noise.

### The one item still outstanding (#6)

`execute()` still nulls `currentElement` before awaiting `request.create()`, so a rejected embed leaves the already-finalized view's `<canvas>` in the DOM with no reference left to clear it. I'd triaged this as "fix if you're already in there" rather than a blocker: the trigger is narrow and the worst case is one leaked canvas rather than the 48 this PR eliminates. Reasonable as a follow-up — just noting it's the one path where the new mechanism can still pin GPU memory.

### Verification (corrected from my first review)

My first review carried a caveat that nothing was backed by a test run. That was true then; it isn't now. CI on db01ea9 is green across the board, and `.github/workflows/test.yml` runs both `npm test` (so the new `chartLifecycle.test.js` / `hostPolling.test.js` cases, including the batched-entries regression, genuinely execute) and the `tests/ui` Playwright suite (so the new lazy-render test executes). `E2E Spaces Test` also passes. I still have not run anything locally.

### Two observations, neither blocking, neither worth a thread

- Opting the initial load out of the timeout removes the watchdog for a genuinely hung first request — recovery now depends on the `$effect` firing `cancel()`. Reasonable trade, just a different failure mode than before.
- `readParquet` only caches on success, so an aborted static download is discarded rather than completing into the cache. Bounded in practice, but it makes `cancel()` more expensive on static dashboards than on live ones.

Nice turnaround. The batched-entries test and the `plots.last` assertion are both better than what I proposed.

PR #686, opened 2026-09-07
Responsive auto layout for charts, sidebar, and tabs

pngwn:
"Plots per row" was fixed at 4, so a phone got four unreadable slivers. Adds an `Auto` default that sizes plots to the container width — 1 column on mobile, up to 4 on desktop. Fixed 1–6 unchanged.

Also fixes two things that made narrow windows unusable: the 290px sidebar now collapses below 768px (and comes back when widened, unless you've toggled it yourself), and the tab strip scrolls instead of clipping tabs off the right edge.

Worth a look: histograms were pinned to 1/row and now follow the setting; Auto keeps plot widths uniform instead of stretching a lone plot full-width.


abidlabs (approved):
Beautiful, LGTM thanks for the nice PR @pngwn!

https://github.com/user-attachments/assets/c7e2c52e-8988-463f-ab7b-a99db28944ec

Just made a few tweaks related to testing & query params.

